PR #61, opened 2019-03-03
Use a window hook to detect the minimize

Abestanis:
On windows, use a [window hook](https://docs.microsoft.com/windows/desktop/WinAuto/registering-a-hook-function) to detect when the Thunderbird window is minimized instead of a check every 250 milliseconds.
This reduces CPU usage and makes the hiding of the window instant.

### DIFF
--- a/src/windowtools_win.cpp
+++ b/src/windowtools_win.cpp
@@ -1,15 +1,29 @@
 #include "windowtools_win.h"
 #include <tlhelp32.h>
 #include "settings.h"
-#include "utils.h"
+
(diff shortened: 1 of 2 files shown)


Abestanis:
Whoops, the commit message is not accurate, sorry about that.
But the content is still correct.

Abestanis:
Of course. It seems I messed up more than the commit message in that commit.
Btw, thank you for making this project! This is exactly what I was looking for after Thunderbird broke the MinimizeToTray addon.

PR #65, opened 2019-03-04
Fix tray icon still visible after birdtray closes on Windows

Abestanis:
Previously, Birdtray would leave a "ghost" tray icon behind after closing.


Abestanis:
I manually tested it in a virtual machine with Ubuntu 18.04 and saw no change in behaviour.

PR #66, opened 2019-03-05
Add windows installer with the Qt installer framework

Abestanis:
This adds a script that creates a windows installer for BirdTray.
The installer has an option to add an autostart entry for BirdTray.

## WIP
This is not finished yet and I mostly want to get your opinion on this.
I'm using the Qt installer framework, because the project is written with Qt, but in my opinion the installer has several issues:

* No back button during the installation.
* Unable to overwrite/update a previous installation.
* Unable to repair the installation. The maintenance tool in the installation directory also can't do it, even though it gives the option to update/modify on the first page.
* The installation is taking up a lot of space (43 Mb on my machine, when only 21 Mb are technically needed). To be fair, 3 Mb are used for translations, but the rest is mostly used by the maintenance tool, which is a lot for a tool that is only able to uninstall the application.

Currently, `windeployqt` also does not copy the sqlite3 dll to the deploy directory, but that might just be because of my complicated compiler setup (I use [MSYS2](https://www.msys2.org/)).

I'm also using a Python script to automate the creation of the installer. The idea is to be able to run the script from Jenkins and then deploy the installer to Github. If you don't want to use Python, I can rewrite the script in another language.

So, what are your thoughts on this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 build*/
 cmake-build-*/
 src/birdtray.pro.user
+installer/birdTray-*
+installer/**/*.xml
+installer/packages/com.ulduzsoft.birdtray.main/data/**
+installer/packages/com.ulduzsoft.birdtray.main/meta/LICENSE
 
 .idea/*
 !.idea/codeStyles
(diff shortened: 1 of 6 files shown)


gyunaev:
I would prefer NSIS based installer, as it also supports silent installation (important for unattended setups)

Abestanis:
Ok, I will look into that.

PR #68, opened 2019-03-08
Fix opening a command prompt when compiling with cmake on Windows

Abestanis:
This tells cmake to compile BirdTray as a win32 GUI executable, suppressing the command prompt.


gyunaev:
I assume this one is no longer needed, right?

Abestanis:
It still is, let me resolve the conflicts.

Abestanis:
Ok, I resolved the conflicts.
It's the `WIN32` option for the `add_executable` that is causing the desired effect.

PR #71, opened 2019-03-17
Add a windows installer with the NSIS framework.

Abestanis:
## Feature overview
* Multi-user support (install without admin privileges)
* Automatic architecture and version detection from birdtray.exe
* AutoRun/Desktop/StartMenu entry and/or StartMenu folder
* Uninstall of previous version with stopping of a running instance
* Installer/Uninstaller silent mode
* Windows version and architecture check
* Prevents multiple installer instances
* Hides licence page after the user agrees
* Installer/Uninstaller have file information and icon
* Installer "Usage" dialog

## Multi-user support
The installer allows the user to install Birdtray for all users or just for her/him. This allows an installation without requiring admin privileges. If administrator privileges are required, they are requested with an UAC prompt. It is technically possible for the user to install Birdtray twice (once for her/him and once for all users, although I don't know why anybody would do that). The setup is able to handle the two installs and uninstalls the correct one upon uninstall depending on which one the user selects.
InstalledPrograms

## Automatic architecture and version detection
The installer is able to automatically extract architecture and version information from the bundled birdtray.exe. If you update the version in the [resource file](https://github.com/gyunaev/birdtray/blob/master/src/res/birdtray.rc#L3), the file name and file information of the installer and uninstaller will automatically be updated during the next generation of the installer.

## Silent mode
I tested the installer in silent mode for installation and uninstallation for the current and all users. I also tested silent uninstallation while birdtray was still running, as well as silent installation with upgrading from a previous version. Note: Installing/uninstalling for all users still requires admin privileges. In silent mode the installer/uninstaller will not request the permission and fail if not started with admin permissions.

## Windows version and architecture check
At startup the installer checks the windows version and architecture.
* The current minimum required windows version is Windows XP, but that is configurable. If the installer detects an older Windows version, it aborts with an error dialog.
* If the installer contains a 64-bit installation of Birdtray and is started on a 32-bit Windows, it will abort with an error dialog telling the user to download the 32-bit installer.
* If the installer contains a 32-bit installation of Birdtray and detects a 64-bit Windows, it will display an information dialog explaining that there is a 64-bit version available for download. The user is able to ignore the message and continue the installation.

## Notes / Outstanding issues
* Please check the configuration at the start of installer.nsi to ensure that I wrote the information correctly.
  - In particular, you could change the `HELP_URL`/`ABOUT_URL` variables to use `mailto:` link to open the email client, if you prefere.
* The installer is currently only written in english and cannot be easily translated. I plan to create a second pull request and work on that once you accept this one.
* [x] On my machine, `windeployqt` is still unable to pick up the dependency on `libsqlite3-0.dll` and fails to place it in der deployment folder. Could you please test if that is also the case on your machine?
* The script uses the [NSISMultiuserPlugin](https://github.com/Drizin/NsisMultiUser). Since it is only available as a Github repository, I copied the relevant files in `install/nsisExtras` with the intention to reduce complexity when automatically generating the installer on a CI server.
* [x] I'm still using a Python script to coordinate the creation of the installer, but at this point it could be replaced with a few shell commands, if you prefer that.
* I hope you notice and like the images I created for the header and sidebar of the installer/uninstaller. I created them using these two images: [Bird](https://svgsilh.com/image/1297891.html) and [Mail](https://svgsilh.com/ffffff/image/1132701.html). They are both placed under the public domain (CC0), so there is no licensing problem using them. If you don't like the header and sidebar, you can of course replace them with your own images, or if you want to change them a bit, I can give them to you in Gimp's XCF-format, so you can easily change the individual layers.
* The installer copies the license into the installation folder and creates a link to it from the start menu. Since the license file has no extension, Windows displays an unknown file icon and can not open the file directly. You should consider adding a `.txt` extension to the LICENSE file.
* Birdtray leaves the tray icon behind when killed by the installer. We should investigate if we can have a signal handler to handle the quit request and remove the tray icon.
* [x] ~There is support for an `/uninstall` flag for the installer, that is added by the NSISMultiUser plugin. At the moment it does not work though. I will try to fix that before finalizing the pull request.~ Fixed.
* [x] There is a warning when compiling the NSIS script: ```warning: 6020: Uninstaller script code found but WriteUninstaller never used - no uninstaller will be created.``` It can be safely ignored, but I want to remove it. It reports one page being included for the uninstaller, but as far as I can see they are all guarded by ```!ifdef UNINSTALL_BUILDER``` guards. Maybe you can find something during the review.
* [x] ~The running Birdtray instance detection is not working when an instance of Birdtray is running with admin privileges. I need to fix this before this PR is ready to be merged.~ The installer now correctly detects if it fails to close the Birdtray process. This only happens, if the user installs Birdtray only for her/him, but manually starts Birdtray with admin privileges. The user can then close the running instance manually. If the installer is started with admin privileges, it can always close Birdtray.

## Images
<details>
<summary>Image gallery showing the install and uninstall process.</summary>

## Installation
InstallerWelcome
InstallerLicense
InstallerUserChoose
If the user chooses installation for all users:
InstallerUserChoose2
On clicking next, the UAC pops up and asks the user for admin credentials.
InstallerPath
InstallerComponents
The next page only shows up if the user selects the "Program Group Entry" section.
InstallerStartMenu
At this point, the installation begins, any running instances of Birdtray are closed and if a previous version exists, it is being uninstalled.
Install last

## Uninstallation
UninstallComponents
The user is asked to confirm the path and is warned, that proceeding will delete files from the directory.
UninstallPath
UninstallFinish

## Usage
If the setup or uninstaller is started with `/?`, it displays this Usage dialog:
Setup Usage
</details>


gyunaev:
windeployqt only copies Qt libraries and *their* dependencies, its not dependency walker for binaries, so your code copying SQLite libraries is correct. See https://doc.qt.io/qt-5/windows-deployment.html:

> The application may require additional 3rd-party libraries (for example, database libraries), which are not taken into account by windeployqt.

Abestanis:
> 1. I'm unsure the whole .idea directory is needed in the project, do you think it is necessary?

It is definitely not needed, but it allows to share configuration for everybody using a Jetbrains IDE and for easy importing of the project.
For example, adding those words to the .idea/dictionaries/Project.xml prevents Clion from reporting them as typos for everybody editing the project with Clion. Which at the moment is probably just me, so if you want me to remove it, I can do that. Doing so won't prevent other devs using Clion to open the project, they'll just have to configure it themselves.

> 2. Is buildInstaller.py depend on Python 2 or Python 3? I'm also unsure Python is needed there - looks like a simple .BAT file should suffice, no? This would simplify builds. You can require all arguments to be present.

At the moment Python 3. As I wrote in the `Notes` section, replacing the script is definitely possible. I will try to do so.

> windeployqt only copies Qt libraries and _their_ dependencies, its not dependency walker for binaries, so your code copying SQLite libraries is correct. See https://doc.qt.io/qt-5/windows-deployment.html:
> 
> > The application may require additional 3rd-party libraries (for example, database libraries), which are not taken into account by windeployqt.

Ok, thank you, I missed that section. The sqlite libraries are currently not copied by the script, so I will add that to the .bat file.

Abestanis:
Ok, I rewrote the python script as a .bat file as you requested.
Please check again that the configuration at the start of installer.nsi is correct.
I would also like to add a `.txt` extension to the LICENSE before you merge this, because at the moment the start menu folder looks like this:
Start Menu folder
If the user clicks on the link, a "Choose program" dialog opens, before they can view the license.

Abestanis:
The windeployqt is placing some image related dlls in `installer\winDeploy\imageformats`, which is adding about 1.5 Mb to the installation. I think the whole folder could be removed, since we only use images for the tray icon and we [only allow png files](https://github.com/gyunaev/birdtray/blob/master/src/dialogsettings.cpp#L430). Support for those seems to be pre built into the main QtGui dll. Is my reasoning correct or did I miss a place where qt would need support for different image formats?

PR #76, opened 2019-03-28
Add event filter to handle WM_CLOSE events on Windows

Abestanis:
This prevents a "ghost" system tray icon when birdtray is killed by `taskkill` or by the installer.

Would you prefer me to put the BirdtrayEventFilter class in a separate new file?


gyunaev:
Please move it to a separate class, thanks!

Abestanis:
Ok, done.

PR #78, opened 2019-04-02
Fix displaying an error dialog when Thunderbird updates on startup, closes #67

### LINKED ISSUE #67: [Windows] Birdtray fails to start Thunderbird when Thunderbird updates
On Windows, when BirdTray starts up and attempts to start up Thunderbird, it displays an error when Thunderbird is applying an upgrade.
BirdTray displays an error dialog stating:
```
Cannot start Thunderbird
Error starting Thunderbird as "...":
Process failed to start: The process cannot access the file because it is being used by another process.
```
Thunderbird will eventually start after applying the upgrade. BirdTray will not recognize that, until the error dialog is closed.

Abestanis:
See #67 for a detailed description of the problem.

## TODO List:

* [x] Implement ProcessHandle for Windows.
* [x] ~Implement ProcessHandle for Linux.~ Not needed
* [x] ~Implement ProcessHandle for Mac.~ Not needed


gyunaev:
Mac is same as Linux, process is not locked when updating, so no need to do anything specific there.

Abestanis:
Thank you for enabling Travis. 👍 😄 
I was waiting for your confirmation in #67 for this pull request before testing it, but I'm also assuming it's the same as Linux.

PR #80, opened 2019-04-07
Add continuous deployment for Windows

Abestanis:
This automatically creates a new release on Github with the two installers for Windows, if you push a tag to the master branch.

## TODO for you after you merge the pull request
* Enable travis for this repository again, so it runs for all every commit made for all branches.
* Replace the [deploy key](https://github.com/gyunaev/birdtray/compare/master...Abestanis:feature/winContinuousDeployment?expand=1#diff-354f30a63fb0907d4ad57269548329e3R155) with a key for this repository as described [here](https://docs.travis-ci.com/user/deployment/releases/#authenticating-with-an-oauth-token).
* Increase the version number in [the resource file](https://github.com/gyunaev/birdtray/blob/master/src/res/birdtray.rc) and push a tag with the version number, to test the whole thing.

## Further ideas
* I could also write a script that automatically updates the version in the resource file based on the tag. 
It would make the change before compilation and commit the change after the deployment. Do you want that?


gyunaev:
I would rather not have automatic continuous deployment. Doing CI builds is valuable, but I prefer to have releases on controlled schedule. Otherwise those random builds end up spread all over Internet, and people come and submit issues from those nighty builds which were never intended to be released. So if you can please remove the deployment part, this would be great.

Abestanis:
I think I might have used the term "Continuous Deployment" wrong, sorry about that. I agree that we don't need nightly builds for every commit we push.

That's why the deployment is [configured](https://github.com/gyunaev/birdtray/pull/80/files#diff-354f30a63fb0907d4ad57269548329e3R160) to run only if you push a **Git tag** to the repository.

That behaviour matches what you have been doing manually [on Github](https://github.com/gyunaev/birdtray/releases) for every release. So you see, you still remain in full control of when a new release is to be deployed. But instead of creating the commit manually on Github, you will have to push a tag, wait for the deployment and then possibly add a changelog.

This can be configured even further. For example:
* At the moment, the deployment will create a release on Github with the message of the last commit. You can then edit that message on Github, to include a changelog. But we can also configure the deployment to only create a draft release, so you can edit the message and then "publish" the release to Github. This would also prevent an accidental full automatic release, as you'll need to submit the release every time.
* Right now, any tag that is pushed causes a deployment. We could filter the tags, so only tags with the format `x.y.z` trigger a deployment.
* If I understand [the docs](https://docs-staging.travis-ci.com/user/deployment/#conditional-releases-with-on) correctly, a tag on any branch will cause a deployment. I'm sure we could filter that as well to only trigger the deployment for tags on the master branch.

All in all, I think the deployment will simplify creating releases for you, which havethe Windows installers attached. Let me know If you still want me to remove the deployment or if you want me to add any of the configuration I described.

gyunaev:
I'd remove the deployment part by now, as I assume it doesn't include testing, and I generally test the stuff manually when releasing it.

Abestanis:
Yes, we do not have any automated tests, so there is no testing performed, just compiling of the source code.
I will comemnt out the deployment config, so you can easily add it later once we have some tests.

As a note though, you could do all your testing before you push the tag. Additionally, if you want to test the Windows binaries generated during deployment, you could change the config as discussed above to only crate a draft release, get the binaries from the draft release, test them and then submit the release.

Abestanis:
It looks like you might have forgotten to enable travis for this repository again? I just created a pull request and I don't see the travis indicator.

PR #81, opened 2019-04-08
Fix a crash when trying to clear accounts when we have no accounts

Abestanis:
If a user had no accounts configured for monitoring and switched the method to parse unread notifications from Mork to Sqlite, we would display a dialog asking them to delete all old accounts and then crash, after the user clicked continue.
These changes ensure that we don't crash and that we don't show the dialog if we don't have any accounts to delete.


Abestanis:
I saw that function and noticed it takes an argument. I didn't realize that there was a default value for the parameter though, so I will remove the other function.

PR #82, opened 2019-04-08
Cleanup sqlite database path handling

Abestanis:
* Prevent appending an additional slash (This is how it was before: `C:/path/to/folder\`).
* Display the path with the native file separators (was always `/` before).
* Centralize the knowledge about `global-messages-db.sqlite` in DatabaseAccounts. This reduces code duplication.


Abestanis:
It will work for existing installations, QDir can handle both `\` and `/` slashes. Old paths will keep the additional slash at the end, but that won't cause problems.

PR #86, opened 2019-05-07
Fix compilation for Qt5.5

Abestanis:
Looks like older versions of Qt (<= 5.5) don't like connecting QApplication::quit directly to a menu action.

See https://travis-ci.org/gyunaev/birdtray/jobs/521692228


Abestanis:
Hm, @gyunaev it looks like Travis is executing the builds for the commits now, but doesn't send the result to Github. Maybe it is related to this? https://github.com/travis-ci/travis-ci/issues/9440

PR #94, opened 2019-06-10
Make it harder to accidentally delete files on uninstall

Abestanis:
This will delete only the files that were installed during installation. If the user manages to save another file into the installation directory of Birdtray, the uninstaller won't delete it.

The installer also now handles replacing an existing installation, if the uninstaller is missing.


Abestanis:
@gyunaev Please enable the Travis integration for this repository. Travis is running for the pull requests, but  the results are not reported back to Github.

PR #99, opened 2019-06-18
Prevent crash from font boldness exceeding 99

Abestanis:
A value greater than 99 will cause a crash.
See https://doc.qt.io/archives/qt-4.8/qfont.html#Weight-enum

### DIFF
--- a/src/dialogsettings.cpp
+++ b/src/dialogsettings.cpp
@@ -146,7 +146,7 @@ void DialogSettings::accept()
     pSettings->mThunderbirdCmdLine = leThunderbirdBinary->text();
     pSettings->mThunderbirdWindowMatch = leThunderbirdWindowMatch->text();
     pSettings->mHideWhenMinimized = boxHideWhenMinimized->isChecked();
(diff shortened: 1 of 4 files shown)


Abestanis:
The reason I did it this way is so that it is impossible to use `mNotificationFontWeight` anywhere else in the code in the future and make the same mistake. How about we make `mNotificationFontWeight` private and create a getter with the `qMin`?

The scenario you've described above is impossible, because it would require the user to have selected a value greater than `99` (`198%`) before. This is not possible, because Birdtrayy would then have crashed.

Also, I can't find any indication that the Qt-Team will change the restriction to a value higher than `99` in a future version of Qt. The [current docs](https://doc.qt.io/qt-5/qfont.html#Weight-enum) still mention the upper bound of `99`.

But I do think that the maximum of `198%` does not look good. How about we replace the number input with a slider instead?

gyunaev:
Sounds reasonable. Let's do it your way then. 

Slider would take more space in dialog, and its quite small piece of functionality.

PR #119, opened 2019-09-08
Added portable save of settings in file for Windows, when birdtray.ini file exist

thunder2:


thunder2:
I use a function to create a QSettings instance because there are two callers (Settings::load and Settings::save).
When you use "QSettings&" as return value and return a local QSettings object, the settings object will be destroyed after the exit of the function and the reference pointer in the callee function will be invalid.

```
static QSettings& createSettings()
{
    QSettings settings(...);
    return settings;
}

QSettings &settings = createSettings();
settings --> invalid
```

Something like this works, but then you have a global QSettings instance for the whole runtime.
```
static QSettings &createSettings()
{
#ifdef Q_OS_WIN
    QFileInfo applicationFilePath(qApp->applicationFilePath());
    QFileInfo fileInfo( QDir( qApp->applicationDirPath() ), QString( "%1.ini" ).arg( applicationFilePath.baseName() ) );

    if ( QFileInfo::exists(fileInfo.absoluteFilePath()) )
    {
        static QSettings settings( fileInfo.absoluteFilePath(), QSettings::IniFormat );
        return settings;
    }
#endif

    static QSettings settings;
    return settings;
}

void Settings::save()
{
    QSettings &settings = createSettings();

```
Then I prefer a protected member variable "QSettings*" on the class Settings initialized once in the constructor Settings::Settings to avoid checking the file for every call to Setting::load or Setting::save.

gyunaev:
In this case you can use QSettings member in Settings. Since Settings is instantiated a single object,  you can just create QSettings member of it, and use it through. Just call sync() after setting the settings in save().

thunder2:
I prefer PR #120

gyunaev:
Merged #120 instead.

PR #124, opened 2019-09-17
Fix: MorkParser crashes on `\\\n\\)` (5C 0A 5C 29), unread count incorrect

Tcc100:
The MorkParser crashes for some of my msf files.
This makes the counter not update for these files (exception is thrown) and generates a SIGABRT when `--dump-mork` is used.

I am not sure why that `if` was added in the first place and why another character is skipped (line 351), but it cannot parse an escaped newline followed by another escaped character.

### DIFF
--- a/src/morkparser.cpp
+++ b/src/morkparser.cpp
@@ -342,13 +342,8 @@ void MorkParser::parseCell()
 			break;
 		case '\\':
 			{
-				// Get next two chars
-				char NextChar= nextChar();
-				if ( '\r' != NextChar && '\n' != NextChar )
-				{
-					Text += NextChar;
-				}
-				else nextChar();
+				// Get next char
+				Text += nextChar();
 			}
 			break;
 		case '$':


Tcc100:
No, as you can see in the hex, the newline is in raw in the file, preceded by a `\`. It's not `\n`

gyunaev:
> No, as you can see in the hex, the newline is in raw in the file, preceded by a `\`. It's not `\n`

Yes, I see it in your dump, but I still don't understand why you removed the original handling of \n and \r (which are skipped in original, but in your case are converted to n)?

To start with, can you please delete/move away your Mork file, and click on "Properties" -> "Repair folder" in Thunderbird? I'd like to exclude the situation where Mork file is simply corrupted, as I'm not going to add support for corrupted morks - this format is not designed to detect or handle corruption, and in those cases I'd rather crash.

Tcc100:
I don't see any reason to strip newline characters from these strings, they are just part of the content.

Nevertheless if you want to skip newlines, you have to remove the else part (line 351), the char is already skipped and this line also skips any following character.

gyunaev:
Have you tried to repair the folder, did the generated Mork have the same issue?

Tcc100:
Sorry, repairing and testing took some time.
My new msf does not contain `5c 0a 5c` anymore, which means that the original `dump-mork` does not crash.
But if you read the dump-mork and compare it to the hexdump of the file, you should find a character missing after every escaping sequence.

gyunaev:
This means there is no need to change the parser. This format is undocumented, so there is no definition of what is proper behavior here. I'd rather leave the parser as-is.

Tcc100:
The format is documented: https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Mork/Structure#Literals
You are right that line breaks are to be omitted after `\`, but that `else nextChar();` is definitely wrong and will continue to cause problems for some people.

PR #133, opened 2019-10-15
Update checker and auto updater

Abestanis:
This adds an update checker that checks for a new version of Birdtray. It can be run manually through the settings or configured to run automatically when Birdtray starts. If a new update is found, a dialog is displayed showing the release notes and a download button.

On non-Windows platforms, a click on the download button opens the [releases page](https://github.com/gyunaev/birdtray/releases).
On Windows, if found the correct installer will be downloaded. Once the installer is downloaded and the user confirms the update, the installer is started and Birdtray closes.

### Technical details
The information about the new releases are fetched via the [Github releases api](https://developer.github.com/v3/repos/releases/). As such, the update checker will only detect a new update, once a Github release is created. The update checker also requires the git tag, with which the release was created, to match the pattern [here](https://github.com/gyunaev/birdtray/compare/master...Abestanis:feature/updater?expand=1#diff-ebef2cd49df15ec9777d8fa2ed5aed3eR24) in order to parse the new version number and determine, if it is newer than the current Birdtray version. This requires you to respect the pattern when creating the git tag, otherwise the version will not be recognized.

### Discussion
* I enabled the update checker on startup by default. Is that ok with you?
* This makes Birdtray require the Qt network module (and the winextras module on Windows). Should I change the install instructions in the README to reflect that?

### Images
<details>
  <summary>Images of the new dialogues</summary>
  
  Note: These images show how the update dialogs would have looked like for the last update.
  * The new setting in the advanced section:
  settings
  * The dialog that is displayed, when a new update is found (on Windows):
  newUpdateDialog
   * The same dialog on Ubuntu:
   ubuntu
   * The download dialog:
   download
   * The update and restart confirm dialog:
   restart
</details>

### DIFF
--- a/src/autoupdater.cpp
+++ b/src/autoupdater.cpp
@@ -42,6 +42,10 @@ AutoUpdater::~AutoUpdater() {
         downloadProcessDialog->deleteLater();
         downloadProcessDialog = nullptr;
     }
+    if (installerFile.isOpen()) {
+        // We quit in the middle of downloading
+        installerFile.remove();
+    }
 }
 
 void AutoUpdater::checkForUpdates() {
--- a/src/autoupdater.cpp
+++ b/src/autoupdater.cpp
@@ -58,6 +58,14 @@ void AutoUpdater::onRequestFinished(QNetworkReply* result) {
         if (isVersionInfoResult) {
             emit onCheckUpdateFinished(result->errorString());
         } else {
+            QString errorMessage;
+            if (installerFile.error() != QFileDevice::NoError) {
+                errorMessage = tr("Failed to save the Birdtray installer:\n")
+                               + installerFile.errorString();
+            } else {
+                errorMessage = tr("Failed to download the Birdtray installer:\n")
+                        + result->errorString();
+            }
             installerFile.remove();
             bool wasCanceled = false;
             if (downloadProcessDialog != nullptr) {
@@ -67,8 +75,7 @@ void AutoUpdater::onRequestFinished(QNetworkReply* result) {
                 downloadProcessDialog = nullptr;
             }
             if (!wasCanceled && QMessageBox::critical(
-                    nullptr, tr("Installer download failed"),
-                    tr("Failed to download the Birdtray installer:\n") + result->errorString(),
+                    nullptr, tr("Installer download failed"), errorMessage,
                     QMessageBox::StandardButton::Retry | QMessageBox::StandardButton::Cancel)
                                 == QMessageBox::StandardButton::Retry) {
                 startDownload();
@@ -113,6 +120,7 @@ void AutoUpdater::startDownload() {
         return;
     }
     if (haveActualInstallerDownloadUrl) {
+        installerFile.unsetError();
         if (installerFile.isOpen()) {
             installerFile.reset();
         } else {
@@ -207,8 +215,18 @@ void AutoUpdater::onInstallerDownloadFinished(QNetworkReply* result) {
                     tr("Failed to download the Birdtray installer:\nInvalid redirect: ")
                     + redirectionTarget.toString(), QMessageBox::StandardButton::Abort);
         }
+    } else if (installerFile.write(result->readAll()) == -1) {
+        QString errorMessage(tr("Failed to save the Birdtray installer:\n")
+                           + installerFile.errorString());
+        installerFile.remove();
+        if (QMessageBox::critical(
+                nullptr, tr("Installer download failed"), errorMessage,
+                QMessageBox::StandardButton::Retry | QMessageBox::StandardButton::Cancel)
+                == QMessageBox::StandardButton::Retry) {
+            startDownload();
+            return;
+        }
     } else {
-        installerFile.write(result->readAll());
         installerFile.close();
         if (downloadProcessDialog != nullptr && !downloadProcessDialog->wasCanceled()) {
             downloadProcessDialog->onDownloadComplete();
@@ -236,10 +254,9 @@ void AutoUpdater::onInstallerDownloadFinished(QNetworkReply* result) {
 void AutoUpdater::onDownloadProgress(QNetworkReply* result, qint64 bytesReceived,
                                      qint64 bytesTotal) {
     if (downloadProcessDialog != nullptr) {
-        if (downloadProcessDialog->wasCanceled()) {
+        if (downloadProcessDialog->wasCanceled() || installerFile.write(result->readAll()) == -1) {
             result->close();
         } else {
-            installerFile.write(result->readAll());
             downloadProcessDialog->onDownloadProgress(bytesReceived, bytesTotal);
         }
     }
--- a/src/autoupdater.cpp
+++ b/src/autoupdater.cpp
@@ -14,7 +14,6 @@
 #include <QtWidgets/QGridLayout>
 #include <QtWidgets/QLabel>
 #include <QtWidgets/QMessageBox>
-#include <QFile>
 #include <QDir>
 #include <QApplication>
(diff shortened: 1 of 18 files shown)


Abestanis:
The Travis build failed because of #134

gyunaev:
Honestly I'm not fan of built-in autoupdaters because of security issues. 

All it takes in its current form a simple DNS hijacking, and this updater would install malware. At minimum the update/installer should be digitally signed, and its signature verified; but the complexity of this is unnecessary.

Checking for the new update and popping up the notification dialog is fine, but it should not be on by default, and the dialog should have an option to "never show me this again".

Abestanis:
"Simple" DNS hijacking is actually not that big of an issue, because all api requests and downloads are done via https. An attacker would have to find a way to create a fake github.com ssl certificate, and the only way to do that would be to get the private ssl key from Github, which is near impossible to do. Additionally, because we are using Github as our download server, which is owned by Microsoft, we can be sure that there are a lot of defence systems to protect the private ssl key of Github as well as quickly handle any DNS hijacks, if they would ever occur.

Also, I know you are using a Linux system so you are probably used to the system package manager to update programs for you. On Windows, we don't have that, which is why this is such a nice thing to have and why a lot of other programs offer this functionality (e.g. Git for Windows, VLC...).

I will disable the update check by default, but I don't think a "never show me this again"-button makes sense. You can just disable the automatic check to never see the dialog. How about a "don't remind me again for this version"-button, that will stop the notification until the next version is released?

gyunaev:
Indeed with HTTPS it would take more than DNS hijacking. I'm not sure what exactly HTTPS handler in Qt does when it encounters invalid certificate, but I'm pretty sure it will not simply ignore it; likely would result in error. It is possible to generate a valid github.com certificate without its private key, but this is not in scope for the average hacker, more of a nation-state.

The check would require external connectivity and thus should be explicitly authorized by the user. You can have an option in the installer asking for this, or simply leave it disabled. If you think its necessary for Windows, okay. As long as it is not present in the non-Windows version, fine with me as I myself don't care about Windows at all :)

gyunaev:
Could this code be easily excluded on Linux, so it is Windows-only?

gyunaev:
A few other things to check for you on Windows:

- Graceful app shutdown when the app is closed by user while a check for update is initiated;
- Handling of SSL errors and invalid replies (such as WiFi captive portals which redirect all traffic to itself) - what happens in onReleaseInfoRequestFinished() if you got some HTML page with "Click on this button to connect to WiFi" instead of JSON?;
- Proper handling of offline machine. I'm not sure about how it works on Windows, but if I boot my laptop with no Internet, does every program I start pop up a message box telling me it cannot say if new version is available? Does it also do so every 180 seconds or how often does it check?
- "installerFile.write(result->readAll());" will read 40Mb into memory buffer, and then write it to disk in one piece. And the user machine will freeze, blocked by disk I/O. This is not good user experience, blocking user for aux functionality. This is also happening on UI thread. Up to you if you wanna change this. And error handling here may also be useful (do you care if disk is full and write failed?)
- You add a new connection each time a downloadDialog is shown; connect probably only makes sense once, when you `new` this dialog.

Abestanis:
I added an `Ignore this version` button as described above:
image

The download code can only be called on Windows because [this condition](https://github.com/gyunaev/birdtray/pull/133/files#diff-ebef2cd49df15ec9777d8fa2ed5aed3eR113) is only set [here](https://github.com/gyunaev/birdtray/pull/133/files#diff-ebef2cd49df15ec9777d8fa2ed5aed3eR172), when [this function](https://github.com/gyunaev/birdtray/pull/133/files#diff-ebef2cd49df15ec9777d8fa2ed5aed3eR130) does not return -1, which it can only do on Windows. But I can add an `#ifdef Q_OS_WIN` around the download condition, if you want.

gyunaev:
> The download code can only be called on Windows 

I'd honestly prefer this whole code (and dependency on QtNetwork) to be only compiled/linked on Windows. But probably this would require too much effort.

Abestanis:
Thanks for reviewing this, I will have a look at it tomorrow.

> "installerFile.write(result->readAll());" will read 40Mb into memory buffer, and then write it to disk in one piece. And the user machine will freeze, blocked by disk I/O.

I've searched online for a better way of doing this, but found nothing so far. I will keep looking.

> I'd honestly prefer this whole code (and dependency on QtNetwork) to be only compiled/linked on Windows.

The update check (not the update) does require the QtNetwork dependency as well.

gyunaev:
> I've searched online for a better way of doing this, but found nothing so far. I will keep looking.

You save the content to disk in the progress callback. This will also help if your update suddenly gets really huge (i.e. 500+Mb) - you don't want to keep it in RAM. This will also let you support HTTP resume. Some people have unreliable Internet, and cannot download 50Mb in one session.

> The update check (not the update) does require the QtNetwork dependency as well.

I know.

Abestanis:
> Graceful app shutdown when the app is closed by user while a check for update ...

I tested quitting in all possible situations and was always able to do so. If the user quits while downloading the installer, the installer file will be deleted.

> Handling of SSL errors ...

I used my hosts file to redirected requests to the Github api to Stack Overflow. That should have the same effect as the DNS hijacking that you mentioned:
Error dialog on SSL error

> ... and invalid replies

It turns out the `QJsonDocument::fromJson` just returns an empty json document, when it gets anything that it can't parse as json. Because there is no `tag_name` in that json document, nothing happens at all; we behave as if we didn't find a new version. I think this is an acceptable behaviour.

> Proper handling of offline machine.

The automatic update check never shows an error dialog, if it fails. Only the manual check from the settings displays an error dialog on failure. The automatic check is retried when the user connects to a network, until it succeeds.

> "installerFile.write(result->readAll());"

This has been taken care of, as you suggested. Thanks for the hint by the way, I was unaware that I could use `readAll` in the `onDownloadProgress` function to write the currently available downloaded data to the file.

> And error handling here may also be useful (do you care if disk is full and write failed?)

Errors while writing to the installer are now gracefully handled with a retry-abort-dialog.

>  You add a new connection each time a downloadDialog is shown; connect probably only makes sense once, when you new this dialog.

I'm not sure where exactly you are referring to. Could you please create [a comment](https://help.github.com/en/articles/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request) in directly in the code to mark the spot?


I put the auto update in the TrayIcon class, because I needed something to store state, but it doesn't really belong there. Would you have any objections against a pull request in the future, that would create a subclass of `QApplication` that we could use for stuff like that. Doing so would also allow us to get rid of all the globals we currently have (like `pSettings`).

Abestanis:
Travis failed because it couldn't download the sqlite sources. Their download server must have been temporarily down.

src/autoupdater.cpp
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+        if (downloadProcessDialog == nullptr) {
+            downloadProcessDialog = new UpdateDownloadDialog();
+        } else {
+            downloadProcessDialog->reset();
+        }
+        downloadProcessDialog->show();
+        QNetworkReply* reply = networkAccessManager->get(QNetworkRequest(downloadUrl));
+        connect(reply, &QNetworkReply::downloadProgress, this,
```
Abestanis: If you were referring to this connect, each time we create a new get request, a new `QNetworkReply` is generated, so we need to connect to the signal of this new object.

gyunaev: Yeah, right.

Abestanis:
Thanks for merging! And thank you for the good review and finding those remaining problems. 😃

PR #144, opened 2019-10-23
Add an optional border to the unread counter

Abestanis:
This lets the user add a border around the unread mail number in the system tray. This improves readability, because it increases the contrast of the text to the background image. The user can configure the color and width of the border, as well as disable the border entirely.

Without border:  image

With border (white, 30% width): image

We also change the default notification color from radioactive green to blue. I don't think anybody kept the green.
Additionally, I added `Off` and `Fastest` labels to the blink speed slider, because it isn't clear at first sight, what the slider does.

The border is **enabled by default**, is that ok with you?

<details>
  <summary>The new settings</summary>

image
</details>

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -155,8 +155,9 @@ void Settings::load()
             "common/defaultcolor", mNotificationDefaultColor.name() ).toString() );
     mNotificationBorderColor = QColor(mSettings->value(
             BORDER_COLOR_KEY, mNotificationBorderColor.name()).toString());
-    mNotificationBorderWidth = mSettings->value(
-            BORDER_WIDTH_KEY, mNotificationBorderWidth).toUInt();
+    mNotificationBorderWidth = mSettings->value( // Disable border on existing installations
+            BORDER_WIDTH_KEY, mSettings->value("common/defaultcolor").isNull() ?
+                              0 : mNotificationBorderWidth).toUInt();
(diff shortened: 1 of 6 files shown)


gyunaev:
Please make it disabled by default to keep it backward-compatible.

Abestanis:
Do you think the border should not be enabled by default in general, or just to keep it backwards compatible. My personal opinion is that it should be enabled by default for new installations. I could add a simple check, that disables the border on existing installations.

src/settings.cpp
```diff
@@ -147,6 +153,10 @@ void Settings::load()
 
     mNotificationDefaultColor = QColor( mSettings->value(
             "common/defaultcolor", mNotificationDefaultColor.name() ).toString() );
+    mNotificationBorderColor = QColor(mSettings->value(
+            BORDER_COLOR_KEY, mNotificationBorderColor.name()).toString());
+    mNotificationBorderWidth = mSettings->value(
+            BORDER_WIDTH_KEY, mNotificationBorderWidth).toUInt();
```
Abestanis: A check like this:
```suggestion
            BORDER_WIDTH_KEY, mSettings->value("common/defaultcolor").isNull() ? 0 : mNotificationBorderWidth).toUInt();
```

gyunaev: Not on existing installations; for new its fine.

PR #155, opened 2019-11-04
Add translations to Birdtray

Abestanis:
*This has become a monster of a pull request and I do apologise for that.*

This adds the ability to translate the UI of Birdtray in different languages. Translations are stored in `src/translations`. They are split up into two files:

* `main_*.ts`: These are translations of strings that are found in the source code by the `lupdate` tool.
* `dynamic_*.ts`: These are translations of strings, that are not in the source code, but are encountered dynamically (mainly the names of the email folders for now).

To handle locale changes during runtime, the `BirdtrayApp` class was added. This also has the nice side effect removing all our global singletons. I haven't been able to test locale changes though, because on both Windows and Ubuntu you are required to log out in order to apply language changes.

This pull request also adds a german translation, as well as a README entry to show people how to add new and edit existing translations.

For Ubuntu, this adds a new requirement for building Birdtray: `qttools5-dev` for the `lrelease` tool.
This might also be a now requirement for the other Linux distros, but I don't have any computers or vms with those set up, so I can't test that.

The qmake build process makes use of the [undocumented](https://evileg.com/en/post/476/) `file_copies` feature to copy the generated `*.qm` files from the `src\translations` directory to the binary directory. [Support for this was only added in Qt 5.6](https://codereview.qt-project.org/c/qt/qtbase/+/156784), so **this raises the minimal required Qt version from 5.5 to 5.6**. If you find a better way to copy the files, let me know.
Building with cmake does still work for Qt 5.5 though, so if you want I can re-add the Travis check for that version.

The translation files are automatically packaged into the Windows installer and installed to the correct location. I did not add them to the installation when installing via qmake, maybe @ThanosApostolou can add that?


src/birdtrayapp.cpp
```diff
@@ -0,0 +1,148 @@
...
+    installNativeEventFilter(&filter);
+#endif /* Q_OS_WIN */
+    
+    QCommandLineParser parser;
+    parseCmdArguments(parser);
+    
+    QString morkPath = parser.value("dump-mork");
+    if (!morkPath.isEmpty()) {
+        MorkParser::dumpMorkFile(morkPath);
+        exit(1); // TODO: Why 1? Replace with return code of dumpMorkFile
```
Abestanis: This was copied from `main.cpp`. Is there a reason why we return 1 (`EXIT_FAILURE`) and not the 0 returned by `MorkParser::dumpMorkFile`?

src/birdtrayapp.cpp
```diff
@@ -0,0 +1,148 @@
...
+        return;
+    }
+    QString imapString = parser.value("decode");
+    if (!imapString.isEmpty()) {
+        printf("Decoded: %s\n", qPrintable(Utils::decodeIMAPutf7(imapString)));
+        exit(1); // TODO: Why 1? Replace with EXIT_SUCCESS
```
Abestanis: Same question here. Why return `EXIT_FAILURE`?

src/birdtrayapp.cpp
```diff
@@ -0,0 +1,148 @@
...
+                                  const QString &translationName, const QString &path) {
+    if (translator.load(locale, translationName, "_", path)) {
+        return true;
+    }
+    // On Ubuntu, when switching to another language, the LANGUAGE environment variable
+    // does not include the base language, only the language with country,
+    // e.g. LANGUAGE=de_DE:en_US:en
+    // instead of LANGUAGE=de_DE:de:en_US:en
+    // As a result, the translator does not find the <translationName>_de.qm files.
+    // That's why we try to load the translation without the country appendix.
```
Abestanis: This was necessary on my Vm with Ubuntu 18.04 LTS. Can you verify that this also works on your Linux distro?

Abestanis:
If you feel like this pull request is too large, I could split off the `BirdtrayApp` changes into a different pull request.

src/birdtrayapp.cpp
gyunaev: Fine with me.

gyunaev:
It is large, but I see no reasonable way to split it, so let's be it.

src/birdtrayapp.cpp
Abestanis: Sorry, I'm not sure if you meant that you are ok with the current way, or if you are ok with changing these. If it is the first, we should remove the TODOs, if it is the second, do you want me to create the changes or are you on it?

PR #161, opened 2019-11-12
Enable "Warning as errors" when compiling

Abestanis:
This can prevent possible errors which are indicated by the warnings and keeps the code robust.

### DIFF
--- a/src/morkparser.cpp
+++ b/src/morkparser.cpp
@@ -276,13 +276,12 @@ void MorkParser::parseDict()
 			switch ( cur )
 			{
 			case '<':
-				if ( morkData_.mid( morkPos_ - 1, strlen( MorkDictColumnMeta ) ) 
-					== MorkDictColumnMeta )
-				{
-					nowParsing_ = NPColumns;
-					morkPos_ += strlen( MorkDictColumnMeta ) - 1;
-				}
-				break;
+                if (morkData_.mid(morkPos_ - 1, static_cast<int>(strlen(MorkDictColumnMeta)))
+                    == MorkDictColumnMeta) {
+                    nowParsing_ = NPColumns;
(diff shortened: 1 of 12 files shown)


src/morkparser.cpp
```diff
@@ -276,13 +276,12 @@ void MorkParser::parseDict()
 			switch ( cur )
 			{
 			case '<':
-				if ( morkData_.mid( morkPos_ - 1, strlen( MorkDictColumnMeta ) )
```
Abestanis: This file seems to be saved with tabs instead of spaces. Spaces are the standard for the rest of the repository. Should we convert this file to spaces as well?

gyunaev:
The main reason I never enable warnings as errors is that compilers tend to add new warnings, which break compilation for new users who have no idea what to do with it.

Also it requires a number of otherwise useless fixes, which do not add any other value to the app.

Abestanis:
I would argue that these warnings are there for a reason and that they should get fixed.

If you worry about the compiler adding new warnings, we could only enable the "warnings as errors" on Travis. That way we keep the benefit of not having warnings in the codebase, and the user won't get errors when compiling Birdtray.

gyunaev:
This used to be the case before. Now the compilers complain about misaligned (in their mind) tabs-spaces, unused variables, class members initialized in "wrong" order and other crap with zero value.

Travis only is fine, though, as we have control over this environment.

Abestanis:
Ok, should be good now.

Abestanis:
@gyunaev Is this acceptable to you? I have gated the 'warnigs-as-errors' feature behind a cmake flag `-D COMPILER_WARNINGS_AS_ERRORS=ON` which is disabled by default, but enabled for Travis.

src/morkparser.cpp
gyunaev: spaces please.

PR #164, opened 2019-11-14
Improve appdata, qmake install translations system wide

ThanosApostolou:
Hello, I've successfully add birdtray to flathub:
https://github.com/flathub/org.gyunaev.Birdtray
https://flathub.org/apps/details/org.gyunaev.Birdtray

I improved appdata and added a screenshot which will belong to your repository and it will be able to be found by all linux stores/repositories (not just flathub), so I will remove those from flathub if this PR is accepted. I am using a patch to make it work with flathub thundebird, but I will try to see if I can make changes so that the users can easily select the thunderbird execution command by the settings, as it's intended.

I've added the qmake "translations" installation target, for system wide installation, but I wanted to also keep the "translations" directory which is next to the binary because I thought that you might prefer that way (when you run it in your system or whatever). So, I added these ugly `#ifdef` lines in the code cause I couldn't think any better way to do this. The greek translation is empty but I will finish it when I'll have time in the next weeks.

P.S. If you feel that you want to maintain the flathub repository yourself then tell me and we could ask the flathub developers to transfer permissions, otherwise I am happy to maintain this for as long as my free time allows me.


Abestanis:
I have just ported your work on installing Birdtray on Unix systems to cmake in #165. Could you please verify that it still installs all the files in all the right places. Keep in mind that the standard installation prefix is `/usr/local` but can be changed via `-DCMAKE_INSTALL_PREFIX=/usr` (as described in the [updated README](https://github.com/gyunaev/birdtray/pull/165/files#diff-04c6e90faac2675aa89e2176d2eec7d8R62)).

I also think I found a slightly more robust solution for loading the installed translations on linux ([see here](https://github.com/gyunaev/birdtray/pull/165/files#diff-0ddc79b45aa3b9696bbdfd193785e5f2R94)).

About the patch: How about replacing the [macros in settings.cpp](https://github.com/gyunaev/birdtray/blob/master/src/settings.cpp#L12) with a function in [Utils](https://github.com/gyunaev/birdtray/blob/master/src/utils.cpp) like [`getThunderbirdProfilesPaths`](https://github.com/gyunaev/birdtray/blob/master/src/utils.cpp#L210). Then, for Linux you could check if the default Thunderbird location exists and use the flathub Thunderbird location as a fallback.

ThanosApostolou:
@Abestanis Oooo I totally missed that, this will actually make things better. I find cmake much easier to work with. So, I'm closing this in order to test it with both native archlinux packaging and flatpaks and I will ping you too if I need to do cmake changes in another PR.

PR #168, opened 2019-11-15
improvements for appdata and flatpak

ThanosApostolou:
Translations work fine with the cmake installation method both for native packages and flathub packages, so no need for cmake changes.

I added the screenshot to this repository so all linux stores/repositories can find it and not just flathub. So if this is accepted I will remove it from the flathub repo and try to build birdtray using this appdata in the beta channel.

@Abestanis for the execution of thunderbird I use this patch and this script:
https://github.com/flathub/org.gyunaev.Birdtray/blob/master/thunderbird-spawn.patch
https://github.com/flathub/org.gyunaev.Birdtray/blob/master/thunderbird

I think the best option would be , if it's possible, to have a cmake option that I can pass for the default thunderbird command and just pass something like
```
-DTHUNDERBIRD_COMMAND="/usr/bin/flatpak-spawn --host flatpak run org.mozilla.Thunderbird"
```
so that the users can change it to something like 
```
"/usr/bin/flatpak-spawn --host thunderbird"
```
from the options in order to be able to use the native thunderbird (and not the flathub one).

However it seems that the code is looking for just a thunderbird absolute path so changing the settings to a command with spaces and stuff doesn't seem to work. Maybe we can do something better?


Abestanis:
Allowing a command and not just a path will require some additional code changes, but it doesn't look impossible. However, it would be a lot easier if we could just find the Thunderbird executable that is installed by flatpack. I think it will install it in `/var/lib/flatpak/app/org.mozilla.Thunderbird`. I'm not super familiar with flatpak, would it be a bad idea to launch Thunderbird from there, or is that considered bad practice?

ThanosApostolou:
Unfortunately that won't work. Flatpaks are sandboxed and by default they only see their ecosystem. But even if I add access to the Thunderbird directory, then I won't be able to launch it since all the exported commands in flatpaks start with `flatpak run` and of course flatpak isn't available inside the sandbox. If I try to use directly the thunderbird binary instead of the exported command, then it won't launch because it won't be able to find its libs and other stuff.

The recommended way of launching external binaries from a flatpak app (and probably the only way without serious hacking) is the `flatpak-spawn` command which is available inside flatpak sandbox.

Abestanis:
I created a proof-of-concept implementation at #169. Let me know if that works for you.

gyunaev:
Should this one now be closed?

ThanosApostolou:
@gyunaev If you can merge this one too. It includes a screenshot and fixes to appdata which will make your app shown in all linux stores and software centers like discover and gnome-software even outside flathub. It also adds the flatpak thunderbird home folder as candidate to look for profiles. It doesn't conflict with the other PR you merged.

gyunaev:
utils.cpp needs fixing.

Also QStringList Utils::getDefaultThunderbirdCommand() is incorrect. It should not fall back to flatpak just because there is no Thunderbird executable. A user may install Thunderbird later, but the command remembered in config would already be flatpak. Please remove this part.

Abestanis:
That one is from my pull request. @ThanosApostolou Can't we use the presence of the `~/.var/app/org.mozilla.Thunderbird` directory to conditionally fall back to the flatpak command?

gyunaev:
Even this might not be good enough. The problem is that this path is first initialized in Settings::Settings the first time the app is run. Then it is remembered in Settings::save() and then on subsequent app restarts it will load the saved path in Settings::load(). 

This means that whatever is detected as default command during the first start will then be automatically remembered as default command, even if the user installed Thunderbird later. This is why I didn't want any flatpak-related commands hardcoded.

Abestanis:
But this is how we do it now already. We already set the default path whether it exists or not. And if the user chooses to install another Thunderbird version, the current system wouldn't react to it either. That is why we allow to specify the path in the settings, isn't it?

gyunaev:
Yes, but on Linux I'm not aware of any distro which doesn't install it into /usr/bin.

Actually while finally reading the full thread (sorry for that) I like the original idea of making it a cmake command-line option, so whoever builds flatpak version can override it. Do you want me to make it?

ThanosApostolou:
@gyunaev the conflict is fixed. Since PR is mostly about the appdata screenshot, and I just added the path for the profile of the flatpak thunderbird as a candidate, which is useful both for a native birdtray and a faltpak birdtray. So, can we merge this one, so that I can add a beta thunderbird at flathub and do better testing of the future changes?

On a general thought, indeed there are lots of possibilities for the thunderbird command (flatpak, snap, native package, custom user build, etc..) and runtime autodetection doesn't make much sense for it. So, I agree that a cmake option for the default command that maintainers can set during build would be better. If you go this path, it would be for the better to change the "auto detect" button to "reset" or something and make it use the parameter of the cmake for the thunderbird command.

gyunaev:
Ok I'll take care of command line options.

Abestanis:
The "auto detect" button still verifies the existence of the executable and searches it on the PATH if the default location does not exist, so don't change the name.

ThanosApostolou:
Hmmm then maybe since the command can be redetected the current solution as we have it is the best? I don't know, I leave it up to you! Feel free me to ping me for questions/testing or whatever you think I can help with!

PR #169, opened 2019-11-15
Support for starting flatpack Thunderbird

Abestanis:
When installed via flatpak, Thunderbird can not be started with the path to the Thunderbird executable. It has to be started via the `flatpak-spawn` command.

Birdtray will prefer the native Thunderbird executable over the flatpak command, if it is present.

This also changes how the user can edit the Thunderbird command in the `Advanced` settings tab: Instead of entering the command in a text editor, the command and arguments can now be entered in a small dialog with a list view, to clearly separate the arguments from another.

<details>
<summary>Pictures</summary>

The modified "Advanced" settings tab

The Thunderbird command edit dialog
</details>

### DIFF
--- a/src/dialogsettings.cpp
+++ b/src/dialogsettings.cpp
@@ -3,6 +3,7 @@
 #include <QMessageBox>
 #include <QFile>
 #include <QDir>
+#include <QtWidgets/QListView>
 
 #include "dialogsettings.h"
 #include "modelaccounttree.h"
@@ -47,6 +48,8 @@ DialogSettings::DialogSettings( QWidget *parent)
     connect( btnNewEmailEdit, &QPushButton::clicked, this, &DialogSettings::newEmailEdit );
     connect( btnNewEmailDelete, &QPushButton::clicked, this, &DialogSettings::newEmailRemove );
     
+    connect( thunderbirdCommandEditButton, &QToolButton::clicked,
+            this, &DialogSettings::editThunderbirdCommand );
     connect( checkUpdateButton, &QPushButton::clicked,
             this, &DialogSettings::onCheckUpdateButton );
     connect( app->getAutoUpdater(), &AutoUpdater::onCheckUpdateFinished,
@@ -65,7 +68,8 @@ DialogSettings::DialogSettings( QWidget *parent)
     boxHideWhenMinimized->setChecked( settings->mHideWhenMinimized );
     boxMonitorThunderbirdWindow->setChecked( settings->mMonitorThunderbirdWindow );
     boxRestartThunderbird->setChecked( settings->mRestartThunderbird );
-    leThunderbirdBinary->setText(settings->mThunderbirdCmdLine.join(' '));
+    thunderbirdCommandLabel->setText(settings->mThunderbirdCmdLine.join(' '));
+    thunderbirdCommandLabel->setToolTip(settings->mThunderbirdCmdLine.join('\n'));
     leThunderbirdWindowMatch->setText( settings->mThunderbirdWindowMatch  );
     spinMinimumFontSize->setValue( settings->mNotificationMinimumFontSize );
     spinMinimumFontSize->setMaximum( settings->mNotificationMaximumFontSize - 1 );
@@ -95,6 +99,12 @@ DialogSettings::DialogSettings( QWidget *parent)
     mModelNewEmails = new ModelNewEmails( this );
     treeNewEmails->setModel( mModelNewEmails );
     treeNewEmails->setCurrentIndex(mModelNewEmails->index(0, 0));
+    
+    // Advanced tab
+    QStringList thunderbirdCommandLine = settings->mThunderbirdCmdLine;
+    thunderbirdCmdModel = new QStringListModel(thunderbirdCommandLine << "", this);
+    connect(thunderbirdCmdModel, &QAbstractItemModel::dataChanged,
+            this, &DialogSettings::onThunderbirdCommandModelChanged);
 
     // Create the "About" box
     QString origabout = browserAbout->toHtml();
@@ -145,13 +155,12 @@ void DialogSettings::accept()
         }
     }
 
-    if ( boxEnableNewEmail->isChecked() && leThunderbirdBinary->text().isEmpty() )
-    {
-        QMessageBox::critical(nullptr, tr("Empty Thunderbird path"),
+    if (boxEnableNewEmail->isChecked() && thunderbirdCmdModel->rowCount() <= 1) {
+        QMessageBox::critical(nullptr, tr("Empty Thunderbird command"),
                               tr("You have enabled New Email menu, "
-                                 "but you did not specify Thunderbird path."));
-        tabWidget->setCurrentIndex( 0 );
-        leThunderbirdBinary->setFocus();
+                                 "but you did not specify a Thunderbird command."));
+        tabWidget->setCurrentIndex( 4 );
+        thunderbirdCommandEditButton->setFocus();
         return;
     }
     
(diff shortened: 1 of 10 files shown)


ThanosApostolou:
Thx for giving it a shot. I won't be able to test this until Sunday, but I'll let you know when I do.

src/settings.cpp
```diff
@@ -61,6 +54,17 @@ Settings::Settings(bool verboseOutput)
     mUpdateOnStartup = false;
     mUnreadOpacityLevel = 0.75;
     mNewEmailMenuEnabled = false;
+
+#ifdef Q_OS_WIN
+    mThunderbirdCmdLine << R"("%ProgramFiles(x86)%\Mozilla Thunderbird\thunderbird.exe")";
+#else
+    if (QFileInfo("/usr/bin/thunderbird").isExecutable()) {
```
Abestanis: @ThanosApostolou If Birdtray is started via flatpak, are we still able to start Thunderbird if it was not installed via flatpak? Or will it also not find its libraries and data? If so, we need a way to detect if we are installed via flatpak. Do you know best way to detect that?

ThanosApostolou: For birdtray's sandboxed environment, /usr/bin/thunderbird will not be existent. So again we need flatpak-spawn. So the commands are:

  - For flatpak thunderbird
```
flatpak-spawn --host flatpak run org.mozilla.Thunderbird
```
or if you want the exported executable (which does exactly the same)
```
flatpak-spawn --host /var/lib/flatpak/exports/bin/org.mozilla.Thunderbird
```
 - For native thunderbird
```
flatpak-spawn --host thunderbird
```
or if you want full path
```
flatpak-spawn --host /usr/bin/thunderbird
```
 - For snap thunderbird
```
flatpak-spawn --host snap run thunderbird
```


In bash it would be easy to check if a file is existent with
```
flatpak-spawn --host file /usr/bin/thunderbird
```
or even better to find executable thunderbird in any directory in the PATH
```
flatpak-spawn --host sh -c "command -v thunderbird"
```
(`command` is built-in available in shell, that's why we need `sh -c` if we don't want to depend on external tools like `which`).
However I'm not sure how to do this in c++.

gyunaev: Well, I certainly don't want to have this kind of complicated logic in code. Can you just provide preconfigured settings, and put the executable command there?

ThanosApostolou: @gyunaev I agree that these shouldn't be considered in code. What I would like, if it's easily done, is that the execution of thunderbird can be a command with spaces or arguments and not just a binary path. And then just a simple cmake option which would default to `/usr/bin/thunderbird` and I could set it to `flatpak-spawn --host flatpak run org.mozilla.Thunderbird` during build time. Then I could drop the ugly patch and `/app/bin/thunderbird` script that I use in flathub and users could also easily change the command to something else (which now is hard to do and not documented, since they would need to create a separate script file and override flatpak birdtray to have access to that location).

Abestanis: I also agree that there should be no complicated logic in the code, but I'm trying to avoid the cmake configuration parameter at compile time, if we can detect it at run time.
If Birdtray is not able to see the Thunderbird executable when installed via flatpak, then the check above should be suficcient to automatically fall back to the `flatpak-spawn` command without requiring a compile time configuration parameter.

ThanosApostolou: @Abestanis  Ok I tested your branch and it seems to work fine with both flatpak and native birdtray. The only problem I saw was that if I delete all the commands/arguments for launching thunderbird from the settings and birdtray is configured to autostart thunderbird, then birdtray refuses to launch (it closes immediately). I think a good approach would be to detect thunderbird command again, if the list is empty. It would be also nice if you can add a button **reset/clear/autodetect** or whatever which would detect thunderbird command from the start, since this change will break existing flatpak installations when I remove my custom **/app/bin/thunderbird** script.

Abestanis: @ThanosApostolou Thank you for testing.

> ... then birdtray refuses to launch (it closes immediately)

That was actually a bug that caused Birdtray to crash, good catch. We really need some tests...

> I think a good approach would be to detect thunderbird command again, if the list is empty.

I agree with that, if the user saves an empty path we should just set our default path again. Do you agree with this, @gyunaev? This would be a change from the current behaviour. Currently, we allow an empty executable path and just display an error dialog when we fail to launch Thunderbird.

> add a button reset/clear/autodetect

That is also a good idea, we could even run some more extensive checks here.

src/dialogsettings.cpp
```diff
@@ -525,3 +587,24 @@ bool DialogSettings::isMorkParserSelected() const
...
+    QStringList defaultCommand = Utils::getDefaultThunderbirdCommand();
+    if (defaultCommand.count() == 1 && !QFileInfo(defaultCommand[0]).isExecutable()) {
+        return defaultCommand;
+    }
+    QString thunderbirdPath = QStandardPaths::findExecutable("thunderbird");
+    if (!thunderbirdPath.isEmpty()) {
+        return {thunderbirdPath};
+    }
+    if (QApplication::applicationDirPath().contains("flatpak")) { // TODO: Is this the best way to detect if we are installed with flatpak and does it even work?
+        // TODO: Is this really the best way to check for other apps?
```
Abestanis: @ThanosApostolou This is where we could detect that we are installed as a flatpak app and detect the Thunderbird installations. Maybe we can detect them [by their desktop files](https://github.com/flatpak/flatpak/issues/1286#issuecomment-354554684), if we can see the files in our sandbox?
Unfortunately, I don't have the setup to test this, so I would leave this part for you, if that is ok for you. If the tests would become too complicated, we could instead just return the default `{"/usr/bin/flatpak-spawn", "--host", "flatpak", "run", "org.mozilla.Thunderbird"`} here.

gyunaev: Guys, I suggest we simply prepackage the configuration file for flatpaks and not code complex heuristics for executable lookups for this purpose. My understanding is that there would be two flatpacs - one using built-in Thunderbird, and one without Thunderbird. They would come with a different configuration file and start different clients.

I'd rather avoid hard-coding a list of possible thunderbird executables and trying to look them up anywhere in the system. Not only this is wrong from the programming standpoint, but you also would have no guarantee that Thunderbird you've found is the one user actually wants to use.

Abestanis: Is distributing two different versions of Birdtray really the right answer? Or three versions, if you want to support the snap version? I'm not suggesting hardcoding a list of possible locations, but rather search for their app id once we know that we have to call them via the flatpak command.
I agree that we might pick up a different Thunderbird than the user is using (but why have two differen versions installed anyway?), but this component is only ment to autodetect *any* Thunderbird installation.

But if you don't like the added complexity, we can always just return the standard flatpak command.

gyunaev: Distributing the same birdtray executable packaged with two/three different configuration files is a possible answer. Another is wrapping the setup into an installer which would ask the user which Thunderbird they'd like to use, and create the right configuration for Birdtray. This is however something which should be done by the person creating the Flatpak, and not by us.

Ideally I prefer to have no code at all inside the Birdtray which does anything flatpak- or snap-specific. Those are supposed to be application wrappers; it is their job to provide the functionality relevant for wrapping the application.

Abestanis: > ... it is their job to provide the functionality relevant for wrapping the application.

You do have a point there. Ok, let's keep the searching for Thunderbird as simple as possible.

Abestanis:
Ok, I think that this is ready now. @ThanosApostolou let me know, if you are missing anything.
If you do end up needing to deploying Birdtray with two different configurations, have a look at [`Settings::loadInstallerConfiguration`](https://github.com/gyunaev/birdtray/blob/master/src/settings.cpp#L294). You should be able to load the Thunderbird command from there.

ThanosApostolou:
@Abestanis looks good, only one problem. After I click the "auto detect" button it adds the default commands but after I click save if I open the settings again the last parameter `org.mozilla.Thunderbird` has been removed. I haven't looked at the code to see how exactly it works, but it seems like it's missing a new line, since if edit manually the last parameter it leaves an empty line below and it works, but the "auto detect" button doesn't.

Abestanis:
Well, this is embarrassing. When I tested this, it looked good, because if the detected command is just one item, it will get removed, so the list is empty. If the list is empty, we just set it to the standard path. That is why I didn't notice that. Thanks again for testing this.

gyunaev:
Are we good to merge this one too?

ThanosApostolou:
I've tested the latest commits and they work fine, I don't know if @Abestanis wants to add anything else.

Abestanis:
No, all done. Go ahead.

PR #172, opened 2019-11-16
Allow to select the translations to be installed with Birdtray

Abestanis:
Adds the ability to select which translations should get installed from the Windows installer.

Updated component selection page of the installer


Abestanis:
With #176 merged, we are also now able to display readable names for the translations:
Installer page: Birdtray translations selection

PR #173, opened 2019-11-16
Convert tab indentations to space

Abestanis:
As requested in [#161 (comment)](https://github.com/gyunaev/birdtray/pull/161#discussion_r347086177).
I also found tabs in `src/colorbutton.*`.

### DIFF
--- a/src/morkparser.cpp
+++ b/src/morkparser.cpp
@@ -66,8 +66,8 @@ private:
 
 MorkParser::MorkParser( int DefaultScope )
 {
-	initVars();
-	defaultScope_ = DefaultScope;
+    initVars();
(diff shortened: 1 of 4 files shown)


src/morkparser.cpp
```diff
@@ -802,5 +802,5 @@ int MorkParser::dumpMorkFile( const QString& filename )
         printf("\n\n" );
     }
 
-    return 0;
+    return EXIT_SUCCESS;
```
Abestanis: I forgot to commit this in #162 so I'm including it here. `EXIT_SUCCESS` is defined to `0`, so this is **not** a change in behaviour.

Issue #171, opened 2019-11-16
core dump if unable to watch .msf file


pablo-blueoakdb:
**OS**:  openSUSE Tumbleweed / 20191112 (latest)
**Birdtray version**:  1.6.0 
**Thunderbird version**: 68.2.1 (64-bit)
**Birdtray origin**: git pull
**Qt version**: Using Qt version 5.13.1 in /usr/lib64
**Description**
After deleting an .msf file, I did not reflect the change in birdtray.  This caused it to core dump.
**Expected behavior**
birdtray should ignore .msf's that it cannot access.

**To Reproduce**
1. Configure birdtray to monitor a non-existent .msf file
2. Attempt to start birdtray; it core dumps

**Additional context**
I stumbled on the resolution while filing this bug.  I thought, let me delete the .msf file mentioned in the voluminous CLI output:

`UnreadCounter generated an error: Unable to watch /path/to/non-existent/.msf for changes.`

I manually removed the reference in the .config files and now birdtray work.  ;)

Issue #163, opened 2019-11-14
A tooltip with account/folder names that got new mails


morfikov:
Currently when you hover a mouse over the thunderbird icon, no useful information is presented when you get new mail -- the icon just blinks (and optionally it shows a number). Would it be possible to add some tooltip with the account/folder names that got new mails?

saanichtonian:
May I add that, as I work on multiple workspaces, it would be handy to quickly see the 'sender' and 'subject' on a mouseover like the old mail-notification app, or perhaps include a settings page to select what info is displayed on said mouseover. 

(running xubuntu 20.04)

geekley:
Quoting my duplicate issue, example tooltips:
> **Thunderbird**
> example1<span>@</span>gmail.com /INBOX (2)
> example2<span>@</span>gmail.com /[Gmail]/Spam (1)

Or, if you could set custom account names:

> **Thunderbird**
> Personal /INBOX (2)
> Work /[Gmail]/Spam (1)

PR #175, opened 2019-11-16
Ignore missing .msf files

Abestanis:
This was requested in #171.
The drawback here is that the user is not notified about this.

### DIFF
--- a/src/translations/main_de.ts
+++ b/src/translations/main_de.ts
@@ -832,10 +832,6 @@ Wollen Sie fortfahren?</translation>
         <source>Cannot query database.</source>
         <translation>Kann nicht von der Datenbank lesen.</translation>
     </message>
-    <message>
-        <source>Unable to watch %1 for changes.</source>
-        <translation>Kann %1 nicht auf Änderungen überwachen.</translation>
-    </message>
 </context>
 <context>
     <name>UpdateDialog</name>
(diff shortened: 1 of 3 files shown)


gyunaev:
I would keep the original code, but add the QFile::exist check in front of them. The monitoring might not be added because of file watcher-related issues, and user should know about this.

Abestanis:
I will do that for now.

I also had the idea of combining this with the system tray icon tooltip idea from #163.
Instead of all these errors being errors, they would be warnings and they could be for individual mail folders (mork files). We could display a small orange `!` in the icon and display the warning message in the tooltip of the system tray icon. What do you think?

Abestanis:
Something like this:
BirdtrayWarningIdea

gyunaev:
> I also had the idea of combining this with the system tray icon tooltip idea from #163.
> Instead of all these errors being errors, they would be warnings and they could be for individual mail folders (mork files). We could display a small orange `!` in the icon and display the warning message in the tooltip of the system tray icon. What do you think?

I wonder if system tray icon tooltips are supported on all platforms (i.e. Mac)?

Abestanis:
I can test that on Ubuntu and Windows. Since we don't support Mac at the moment, I wouldn't worry about it. Do you like the idea in general?

gyunaev:
> I can test that on Ubuntu and Windows. Since we don't support Mac at the moment, I wouldn't worry about it. Do you like the idea in general?

Yes, its a good idea. I assume that we still print the error in console in this case (especially with -d mode)?

Abestanis:
Awesome, then I will give it a go.

> I assume that we still print the error in console

Yes

gyunaev:
It's ready to be pulled in, right?

Abestanis:
Yes, thanks 👍

PR #179, opened 2019-11-21
updated icons

ThanosApostolou:
@Abestanis I made a little bit bigger the bell and the "B" letter so that they are a bit clearer in 32 and 48 px icons. I removed the thunderbird.png. If you want take the svg and send me the header.bmp or we can include it later in another PR.


src/res/birdtray.svg
```diff
@@ -0,0 +1,172 @@
...
+   width="64"
+   height="64"
+   x="0"
+   y="0"
+   version="1.1"
+   xml:space="preserve"
+   id="svg34"
+   sodipodi:docname="org.gyunaev.Birdtray6.svg"
+   inkscape:version="0.92.4 5da689c313, 2019-01-14"
+   inkscape:export-filename="/home/thanos/Pictures/org.gyunaev.Birdtray6-64.png"
```
Abestanis: Inkscape seems to write a lot of unnecessary stuff into the svg file. I wonder if we should remove that.

Abestanis:
I will change the installer header image once @gyunaev made his decision. I could also have a look on how to export the svg into a multi resolution ico, if you don't want to do that.

gyunaev:
Thank you very much. The icon looks good to me, please go ahead. Can you also please replace all the icons in src/res but keep the file names? They are used on other platforms.

Abestanis:
@gyunaev Should the new icon replace the Thunderbird icon as the default system tray icon?

gyunaev:
> @gyunaev Should the new icon replace the Thunderbird icon as the default system tray icon?

Yes please.

ThanosApostolou:
@Abestanis I added layers 128, 64,48 and 32 to birdtray.ico using gimp, if you can check yourself to tell me if anything is wrong (I'm not familiar with the ico type). I made the B letter just a bit smaller.

@gyunaev I was thinking for the tray icon (which is currently the "thunderbird.png", if I understood correctly) maybe it is better to remove the bell and the letter, so that the mail count look more clean. What do you think?

Abestanis:
@ThanosApostolou The multiresolution icon file looks good, but could you please also include the resolutions for `256x256` and `512x512`? The reason I'm asking is that this icon will be embedded into the Birdtray executable on Windows and if you have the executable or a link to it in an explorer window or at the desktop with the `Large icons` option, it will look wrong.

Alse, there is still a gap between the icon and the border in all of the files. This should get removed. You should be able to do that in Inkscape by following the section `Adjusting the Canvas Size to Fit Your Image` in [this wiki page](https://wiki.inkscape.org/wiki/index.php/Tricks_and_tips).

Abestanis:
@ThanosApostolou Awesome, looks very good. Give me a minute to send you the updated installer header.

Abestanis:
Here you go: [header.zip](https://github.com/gyunaev/birdtray/files/3876691/header.zip)

gyunaev:
Are we ready to merge?

ThanosApostolou:
Yep all finished from my part.

PR #181, opened 2019-11-22
Remove the Sqlite3 parser

Abestanis:
This removes the Sqlite3 parser, which has stopped working for Thunderbird 68+.
This reduces the overall code complexity of the project.

We automatically attempt to migrate accounts which were configured with the Sqlite3 parser to the mork parser. This happens only the first time on launch. If the migration succeeds, an information dialog is displayed. If it fails, a warning is displayed. The migration supports the following uris:
* `scheme://<name>@<address>` will be mapped to `<profileDir>/<schemeFolder>/<address>/**.msf` (`**.msf` = All files with `.msf` ending in folder and subfolders)
* `scheme://<name>@<address>/<name>` will be mapped to  `<profileDir>/<schemeFolder>/<address>/<name>.msf`
* `scheme://<name>@<address>/<path>/<name>` will be mapped to  `<profileDir>/<schemeFolder>/<address>/<path>.sbd/<name>.msf`

The mapping from `scheme` to the appropriate directory `schemeFolder` in the Thunderbird Profile supports:

* `mailbox` -> `Mail`
* Everything else: `<other>` -> `<Other>Mail`, e.g. `imap` becomes `ImapMail`.

These two are the ones I have on my machine. If you know any other schemes that don't work with the *other* rule above, let me know.

I do understand if you don't want to remove the Sqlite3 parser yet. I am just proposing this, because it would make the implementation of the things I am currently working on easier. If you don't want to remove the parser yet, what is your planned timeline on when to remove it?


gyunaev:
I would keep it for now. It is possible that they'd reverse the change in future Thunderbird - I hardly could see any practical reason for keeping the database locked, so may be an error on their side.

rnc:
Is there a Thunderbird ticket verifying what their approach will be? ( or is it'd worth entering one?)

Abestanis:
@gyunaev Ok, makes sense.
@rnc A quick google search did not yield any results. If you want to create a ticket, go ahead. If there is no good reason for them to keep the database locked, then it is probably a bug they don't know about.

PR #182, opened 2019-11-23
Better fix for delete while still running error

Abestanis:
This is a better fix than #174. It could still sometimes crash when Birdtray was exiting.

We no no longer move the `QThread` object to the thread. This causes the destructor tobe executed on the main thread, allowing us to wait for the thread to exit before continuing destruction the object.

I also renamed the source file so it matches the class name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ set(SOURCES
         src/settings.cpp
         src/sqlite_statement.cpp
         src/trayicon.cpp
-        src/unreadcounter.cpp
+        src/unreadmonitor.cpp
         src/utils.cpp
         src/windowtools.cpp
         src/autoupdater.cpp
@@ -84,7 +84,7 @@ set(HEADERS
         src/settings.h
         src/sqlite_statement.h
         src/trayicon.h
-        src/unreadcounter.h
+        src/unreadmonitor.h
         src/utils.h
         src/version.h
         src/windowtools.h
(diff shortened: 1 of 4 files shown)


gyunaev:
There seem to be lots of irrelevant changes which have nothing to do with fixing the issue.

Especially please do not replace 0 with nullptr which requires C++11, while the rest of the app does not.

Abestanis:
> There seem to be lots of irrelevant changes which have nothing to do with fixing the issue.

I thought I use the opportunity to fix some warnings in the file that are displayed in my IDE.

> Especially please do not replace 0 with nullptr which requires C++11, while the rest of the app does not.

That is not true, I used them all over the codebase in the past. Also, what is wrong with requiring C++11? If anybody required on older standard to build Birdtray, they would have said so when we first used `nullptr` about 9 months ago.

gyunaev:
The problem with merging many changes together into single request is that its much harder then to find them out when looking at git log. It is also harder to write changelogs.

If we start requiring C++11, there has to be benefit either for us, or for the user. I don't really see any benefits from this replacement alone. It's basically to handle overloading f( int a ) vs f( void * a), but we don't have any such overloading. I see no other benefits.

Abestanis:
I will remove the unnecessary changes.

As I said, we **already** require C++11 since roughly 9 months.
Apart from modernizing the code, it looks like you need to target C++11 to be able to use the [lambda functions](https://en.cppreference.com/w/cpp/language/lambda), which are really nice and we are already using.

Abestanis:
Oh, I just realized; Do you want me to undo the file renaming too?

gyunaev:
No, file renaming is meaningful, so its okay.

Issue #183, opened 2019-11-23
Current command-line editor in advanced is kinda weird


gyunaev:
The way it is edited is a bit strange and doesn't allow copypasting complex command-lines. @Abestanis what would you say if we split it into two normal edit fields, one for executable (so it is taken as-is with all spaces) and one for the command-line arguments (which go space-separated)?

Abestanis:
The reason I did it that way is to not have to deal with spaces in arguments, e.g `command argument="Arg with space"`.

gyunaev:
I understand, but this is advanced settings tab; people opening it and modifying stuff there are expected to understand what they are doing.

Actually maybe we should use a single "Thunderbird command-line" edit line, and have a small state machine parser to properly convert into QStringList strings like "C:\Program Files\exe\thunderbird.exe" --profile test --title "My test profile"

Abestanis:
Parsing commands is not a trivial task, think about escaped quotes `\"`, single quotes vs double quotes and so forth. If you really want to be able to paste complex command lines, we would have to implement a fully fledged command line parser.
I don't know if that is worth it. In my opinion, you should not need to interact with this dialog very often, so it might be ok if it is not as easy to use as the rest of Birdtray.

Abestanis:
We could of course allow copy paste and just split on all spaces. Joining the parameters back together might be easier than having to enter each of them manually. That way we gain a little usability but don't add much complexity.

gyunaev:
Actually a splitter is quite easy - see 18ad6c7..5f19eb2

Abestanis:
I can't tell if that coveres all the corner cases, since I don't all the details of command line parsing and escaping myself. That is why I would use a libraries for this.

From what I could test, this looks good. A small problem with this implementation is that it writes back the escaped `\`, which we don't want. I think Qt will pass those straight to the process without unescaping them.

If we add this functionality, you should probably extract the dialog from `dialogsettings.cpp` into its own source file and implement it with a `.ui` file. I would then definitely split the command and arguments into two sections.

gyunaev:
We only expect to pass the basic command-lines which we split properly (since we're not using shell). This is not for writing shell scripts :)

PR #184, opened 2019-11-24
Changed advanced command-line editor to use a single input field.

gyunaev:
@Abestanis please take a look if it works for you on Windows. Basically the idea is to have a simple command-line editor allowing advanced users to change entries. This field assumes someone knows what they are doing.


gyunaev:
@ThanosApostolou this also allows specifying a single default command line to start Thunderbird for flatpak/snap:

> 
cmake .. -DOPT_THUBDERBIRD_CMDLINE="/usr/bin/thunderbird --profile 'my best profile'"

Abestanis:
Unfortunately, this doesn't work and I don't think it is specific to Windows.

The simple case (one command without arguments) works fine, even if the command has quotes around it or contains spaces.
Simple arguments (without spaces or quotes) work as well.
But as soon as you add quotes around an arguments containing spaces, the command does no longer work.

This is what I tried to tell you in [#183 (comment)](https://github.com/gyunaev/birdtray/issues/183#issuecomment-557765062), but I didn't explain it very well. Let's see if I can describe it better:

We require to user to surround arguments containing spaces with quotes. A shell would remove the quotes and all characters that are needed to escape other characters. If you type the following into a shell:
```shell script
"C:\Program Files\Python37\python.exe" "C:\Python Scripts\launcher.py" "C:\Program Files (x86)\Mozilla Thunderbird\thunderbird.exe"
```
`python.exe` would receive:
1. `C:\Python Scripts\launcher.py`
2. `C:\Program Files (x86)\Mozilla Thunderbird\thunderbird.exe`

Notice that all quotes are removed before they are passed as arguments.

If I enter the command above into the `Thunderbird command line` field, this is what `python.exe` receives from Birdtray:
1. `"C:\Python Scripts\launcher.py"`
2. `"C:\Program Files (x86)\Mozilla Thunderbird\thunderbird.exe"`

Notice that we did not remove the quotes. As a result, python does not find the script, because a path starting with a quote is not a valid path:
```
C:\Program Files\Python37\python.exe:
can't open file '"C:\Python Scripts\launcher.py"': [Errno 22] Invalid argument
```

Also, since you removed the code that handles escaping of quotes, it is no longer possible to have an argument containing a quote.

gyunaev:
Correct. I've fixed the code and changed the parser logic. Now I strip all quotes when parsing, so the command stored in tbcmdline is ready to launch. I append quotes (in case of spaces) myself.

I could hardly see a case when you need an argument containing a quote.

Abestanis:
Well, how about
```shell script
/usr/bin/thunderbird --profile "my friend's \"best\" profile"
```
In any case, the dialog supported any argument, no matter what.
In my opinion you are sacrificing existing functionality for being able to paste the command.

Abestanis:
What do you think about this:
We could keep the dialog but detect copy&pasting into it. If we detect a copy&paste, we use `Utils::splitCommandLine` to split the pasted string. If we don't split correctly, the user can still edit the arguments manually.

I think this way we get the best of both worlds.

gyunaev:
> Well, how about
> 
> ```shell
> /usr/bin/thunderbird --profile "my friend's \"best\" profile"

I'd say this user wasn't asking for problems, he was basically DEMANDING them.

gyunaev:
> What do you think about this:
> We could keep the dialog but detect copy&pasting into it. If we detect a copy&paste, we use `Utils::splitCommandLine` to split the pasted string. If we don't split correctly, the user can still edit the arguments manually.
> 
> I think this way we get the best of both worlds.

No, we don't. Honestly I didn't like that editor at all. This is not how you edit command line on Linux. There is lots of code. And all this is just to support an abstract use case, which is not even likely to happen. Yes, I don't like it either removing the code someone spent time writing and debugging, but I've been in this business for long enough to see how quickly things could grow unmanageable.

My approach is generally like that: if a user has a totally unique use case, which nobody else is likely to have, I'm not gonna do anything. Those kinds of issues are not project improvement, they are product customization. They have the source; they can fix it ourselves, or hire someone to do it for them (and maintain it).

Abestanis:
Well, that was just my opinion. I also don't see how a little dialog can become unmanageable, if you wouldn't change it again, but I guess I'll have to trust your judgement. Let me quickly verify that the new changes work on Windows.

src/dialogsettings.cpp
```diff
@@ -84,6 +80,21 @@ DialogSettings::DialogSettings( QWidget *parent)
...
     boxShowUnreadCount->setChecked( settings->mShowUnreadEmailCount );
 
+    // Form the proper command-line (with escaped arguments if they contain spaces
+    QString tbcmdline;
+    for ( QString a : settings->mThunderbirdCmdLine )
+    {
+        if ( !tbcmdline.isEmpty() )
+            tbcmdline += ' ';
+
+        if ( a.contains( ' ') )
```
Abestanis: Don't add quotes if they are already present, otherwise you get `""%ProgramFiles(x86)%\Mozilla Thunderbird\thunderbird.exe""` for the default path on Windows.

src/dialogsettings.ui
```diff
@@ -745,7 +745,7 @@
                </sizepolicy>
               </property>
               <property name="text">
-               <string>Thunderbird command:</string>
+               <string>Thunderbird command line:</string>
```
Abestanis: It looks like you forgot to commit the changes in the language files.

src/settings.h
```diff
@@ -132,7 +132,7 @@ class Settings
          * @param arguments The arguments necessary to start Thunderbird.
          * @return The executable that starts Thunderbird.
```
Abestanis: The documentation needs to be updated.

src/trayicon.cpp
```diff
@@ -610,23 +622,24 @@ void TrayIcon::startThunderbird()
...
-    mThunderbirdProcess->start(executable, arguments);
+    mThunderbirdProcess->start(executable, args);
 }
 
 void TrayIcon::tbProcessError(QProcess::ProcessError )
 {
 #ifdef Q_OS_WIN
-    if (mThunderbirdUpdaterProcess->attach() == AttachResult::SUCCESS) {
+    if (mThunderbirdUpdaterProcess->attach() == AttachResult::SUCCESS)
+    {
```
Abestanis: I don't think you wanted to include that.

Abestanis:
You also removed all references to `DialogSettings::searchThunderbird`. Either remove the function or add an `Auto detect` back. It is quite handy to be able to autodetect Thunderbird on the path, so I would prefer to keep it. Also, I think @ThanosApostolou wanted the button.

gyunaev:
Yes, I removed it because this functionality - if implemented at all - should be implemented in the installer, not in the app itself. I have doubts whether it is valuable though. Someone who doesn't know what thunderbird executable is wouldn't go to Advanced (and thus click Autodetect), and someone who does won't need Autodetect.

src/trayicon.cpp
gyunaev: That's windows, right?

src/dialogsettings.ui
gyunaev: I didn't change language files, as I've no idea what to do with translations - I never use/build them.

src/trayicon.cpp
gyunaev: I generally use the next-line brackets, so the functions I modify I do the changes :)

Abestanis:
In my opinion, the assumption that an unexperienced user won't open the `Advanced` tab is quite erroneous. It's not like the tab is hidden or anything. Again, that is just my opinion, though.

src/dialogsettings.ui
Abestanis: The language files are automatically changed when you build the project with cmake if you have the language tools installed.

src/trayicon.cpp
Abestanis: If you can find the time, I would really appreciate it if you add a `.clang-format` file or similar to the project. That way we can both use the same code style for this project.

Abestanis:
Looks good to me. 👍

ThanosApostolou:
Ok, tested it and works fine for my purposes. Maybe since you dropped the "auto detect" button, you can add a "reset advanced options" button at the "Advanced" settings page which resets the options to default values, in case someone messes them and doesn't remember the initial values?

gyunaev:
What would this button do, i.e. what exactly it needs to "reset"? Those fields are user-editable, so users can put proper values there themselves.

ThanosApostolou:
Ok cool, no need to complicate the code since it works fine.

PR #190, opened 2019-11-28
Simplify checking for new and changed mork files

Abestanis:
This is an optimization for the Mork parser and simplification of the logic in `UnreadMonitor::getUnreadCount_Mork`. We no longer read the unread count from all msf files if we add a new msf file. Instead, only the newly added and the changed msf files are read.

### DIFF
--- a/src/unreadmonitor.cpp
+++ b/src/unreadmonitor.cpp
@@ -213,10 +213,7 @@ void UnreadMonitor::getUnreadCount_Mork(int &count, QColor &color)
 {
     Settings* settings = BirdtrayApp::get()->getSettings();
     for (const QString &path : settings->mFolderNotificationColors.keys()) {
-        if (!QFile::exists(path)) {
-            continue;
-        }
-        if (!mDBWatcher.files().contains(path)) {
+        if (!mDBWatcher.files().contains(path) && QFile::exists(path)) {
             if (!mDBWatcher.addPath(path)) {
                 emit error(tr("Unable to watch %1 for changes.").arg(path));
                 continue;
--- a/src/unreadmonitor.cpp
+++ b/src/unreadmonitor.cpp
@@ -212,37 +212,22 @@ void UnreadMonitor::getUnreadCount_SQLite(int &count, QColor &color)
 void UnreadMonitor::getUnreadCount_Mork(int &count, QColor &color)
 {
     Settings* settings = BirdtrayApp::get()->getSettings();
-    bool rescanall = false;
-
-    // We rebuild and rescan the whole map if there is no such path in there, or map is empty (first run)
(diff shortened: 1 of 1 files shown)


gyunaev:
Your change now calls QFile::exists (meaning it calls `stat()`) multiple times a second resulting in disk accesses and thus disk activity - which may be cached or not - all the time. The whole purpose of this initial complexity was to avoid those calls.

Abestanis:
That is true, but I remove the check completely in my other branch where I am working on the warnings as tooltips as discussed in [#175 (comment)](https://github.com/gyunaev/birdtray/pull/175#issuecomment-555689611). I will create that pull request as soon as I verified that it works on both Windows and Ubuntu. For now I will move the condition into the `!mDBWatcher.files().contains(path)` if, so it is not executed every time.

gyunaev:
But what is the purpose of this change? This part hardly needs simplification/optimization, or do I miss anything?

Abestanis:
I also think, we can extract the entire for loop from this function into the settingChanged event and at startup.

This function is potentially called many times and I thought it was wasteful to read every mork file just because we added a new one. I also found to logic confusing, so I wanted to clarify it.

gyunaev:
I don't really think there is anything to optimize here, seriously. Yes, it could have been written better, but its waste of time to optimize a function to remove extra couple CPU cycles. 

If you really wanna optimize, better look at MorkParser itself, its ridiculous.

gyunaev:
Basically any code optimizations make sense if they produce **measurable** performance improvement, or **significantly** (we're talking 3x+) reduce the code size. Otherwise please don't bother. The negative effect of those optimizations is that I would have to spend significantly more time to understand what the new code does.

Abestanis:
Well, I find the new code much easier to read, but I know that is because I wrote it. 😄
Let's close this for now, we should probably focus on tests before we think about performance.

gyunaev:
> Well, I find the new code much easier to read, but I know that is because I wrote it.

This is true. Every developer thinks a certain way, and expects things to be written in a certain way. This is why rewriting code for the purpose of rewriting is discouraged, as in a large team you can easily end up with 20 pull requests rewriting the same piece of code, instead of fixing bugs or adding new functionality.

Abestanis:
In my defence, I think improving the performance or reducing the CPU cycles in this function will be pretty important. At least on my System, Birdtray is running 24/7. Over the course of a year, even a relatively small reduction in CPU cycles spent in this function will have a huge impact on the cycles saved and potential energy saved. At least that is the what I am thinking. But as I said, we can tackle that problem, when we have the time for it.

gyunaev:
Trust me, your DE wastes so many cycles that such improvements are akin to teaspoon of water for a swimming pool.

Issue #189, opened 2019-11-25
Change linux name/id from org.gyunaev.Birdtray to com.ulduzsoft.Birdtray?


ThanosApostolou:
I've submitted the icons, appdata and desktop files needed for linux under the name/id **org.gyunaev.Birdtray** which is in agreement with freedesktop standards. However, I saw that the windows installer is using as company name "UlduZsoft". So, would you want me to change the name/id to **com.ulduzsoft.Birdtray** to all files needed for linux (which is also in agreement with freedesktop standards and flathub) ?

P.S. I've asked the flathub maintainers and I can resubmit the app with the new name, while deprecating the old one so there isn't any problem with that. If you decide to do this change then it should probably happen before a new release, since then it will break the desktop file for those who have already put it in autostart.

gyunaev:
Please check with @Abestanis as he's doing Windows-related stuff.

Abestanis:
Sorry, but I don't think this is really related to Windows. The Windoes installer uses the company name `Ulduzsoft` because of this line in [`birdtrayapp.cpp`](https://github.com/gyunaev/birdtray/blob/master/src/birdtrayapp.cpp#L13) (former `main.cpp`): `QCoreApplication::setOrganizationName("ulduzsoft");`
I think this means you should rename the id, because this line code says Birdtray belongs to the `Ulduzsoft` company. But I don't really know how important the correct naming for the id is and what it will affect. I think @gyunaev must decide on that.

gyunaev:
This is not important for me, but this will affect existing settings retrieval on Linux and Windows, which use this as a storage directory/key.

Abestanis:
I don't think the suggestion is to change that string. From what I understand, every flatpack package has an unique id, and @ThanosApostolou is currently using `org.gyunaev.Birdtray`, but since `Birdtray` is "owned" by a company, the correct id would be `com.ulduzsoft.Birdtray`.

Abestanis:
Changing the id will definitely not affect existing installations for Windows and will only affect those on Linux who installed Birdtray via flatpak with the old id.

Abestanis:
This must be decided before I can create the new release.

gyunaev:
I'd leave it as Ulduzsoft and change other things accordingly.

Abestanis:
Ok, once @ThanosApostolou creates a pull request with the necessary changes, I will create the new release.

PR #193, opened 2019-11-29
Change name/id to com.ulduzsoft.Birdtray

ThanosApostolou:
Resolves https://github.com/gyunaev/birdtray/issues/189

I changed the text object of the svg image to path, because I noticed that the letter changes according to the used font of the system (the exported images like png aren't affected). I also added the svg image to linux installation as a scalable icon.


Abestanis:
Looks good :+1:

> I changed the text object of the svg image to path

Awesome, I run into that problem because the font `FreeMono` is not installed on Windows by default.

PR #195, opened 2019-12-01
Allow to select msf files in mail subfolders

### LINKED ISSUE #194: Allow watching subdirectories
The way my mail directory is set up, my inbox has a bunch of subdirectories where new mails are being moved into based on filters. When new mails appear in those, they won't count towards the number of unread messages in the tray icon when I am telling birdtray to watch the inbox for unread messages, only mails that stay in the inbox directory itself show up. I am currently not able to add the subdirectories separately to count towards it, as they do not appear in the list of directories that can be watched. Only directories in the root appear at all. It would be very nice if you added subdirectories to the list of directories that can be watched, so that I do not miss out on emails that are being sorted into them.

(Info: I am using the latest version compiled from the master branch with Mint Cinnamon 19.2 and Thunderbird 68.2.1)

Abestanis:
This fixes #194.
We allow to select `.msf` files in mail subfolders (folders ending with `.sbd`).
We should merge this before we make a new release.

### DIFF
--- a/src/mailaccountdialog.cpp
+++ b/src/mailaccountdialog.cpp
@@ -1,6 +1,7 @@
+#include <utility>
 #include <QtWidgets/QMessageBox>
 #include <QtWidgets/QFileDialog>
-#include <utility>
+#include <QtCore/QDirIterator>
 #include "mailaccountdialog.h"
 #include "ui_mailaccountdialog.h"
 #include "utils.h"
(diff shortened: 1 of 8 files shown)


Abestanis:
All changes are in areas of the code that are specific to mork files. The dialog is not shown for sqlite accounts.

gyunaev:
I mean, in the UI area would sqlite parser be still functional?

Abestanis:
Yes, the code changes do not affect the sqlite parser at all.
The `MailAccountDialog` is not used when the sqlite parser is enabled ([see here](https://github.com/gyunaev/birdtray/blob/master/src/dialogsettings.cpp#L323)) and the changes in the accounts list display are only reached for paths that end in `.mfs` ([see here](https://github.com/gyunaev/birdtray/blob/master/src/modelaccounttree.cpp#L35)).

PR #196, opened 2019-12-01
Enable Travis autodeploy on tag, but only as a draft

Abestanis:
This is a proposal to enable the automatic creation of a draft release on Github, if a tag is pushed to the master branch.
Building to Birdtray installers for Windows on my machine is not ideal. They should ideally be created in a clean environment.
I know you didn't like the idea in the past. I wanted to restart the discussion and see, if some of the concerns you had might have been resolved now.
A major concern was to be able to manually test Birdtray before creating a release ([comment](https://github.com/gyunaev/birdtray/pull/80#issuecomment-481356568)). You will still be able to do that, as this automation only creates a draft of a release. Once the draft is created, you can download the installer or compile the code and test it. If you find anything, just discard the draft, fix the error and push another commit with a tag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -144,13 +144,14 @@ before_deploy:
       cd ..;
     fi
 
-#deploy:
-#  - provider: releases
-#    api_key:
-#      secure: SECURE_KEY_GOES_HERE
-#    file_glob: true
-#    file: installer/Birdtray-*.exe
-#    skip_cleanup: true
-#    on:
-#      tags: true
-#      branch: master
+deploy:
+  - provider: releases
+    api_key:
+      secure: SECURE_KEY_GOES_HERE
+    file_glob: true
+    file: installer/Birdtray-*.exe
+    skip_cleanup: true
+    draft: true
+    on:
+      tags: true
+      branch: master


gyunaev:
As long as you're taking responsibility for the Windows part, it is fine with me. I have no Windows, so I'm not going to test it.

Abestanis:
Sure, I will test the installer and Birdtray on Windows, before we change the draft to a real release.
When we merge this tough, you need to replace the `SECURE_KEY_GOES_HERE` with the security token.
To get the token, follow the instructions [here](https://docs.travis-ci.com/user/deployment/releases/#authenticating-with-an-oauth-token).

gyunaev:
Why do I need the token? I don't want Travis to push releases directly to Birdtray, they should be uploaded somewhere where they could be tested first.

Abestanis:
They are uploaded to Github as a draft release, meaning that nobody can see the release except the maintainers of Birdtray. But we can easily turn them into a real release with the click of a button, once we finish our tests. Again, nobody but us will be able to see the draft release or the files that are attached, until we hit the `release` button.

Abestanis:
See [here](https://help.github.com/en/github/administering-a-repository/creating-releases) for a description of the release drafts.

gyunaev:
The setup seems to be quite complicated from the security standpoint. I'd pass it for now, the security implications are unknown and the benefits are totally non-obvious to me. I can build those if you want - I have a Windows VM which I use to build windows stuff.

Abestanis:
Let me try to clarify:
1. Security implications:
 * None. Nobody will be able to use that key to access this repository or create a release, nobody will be able to do anything other than before except the Travis server who can decode the key into a Github Oauth2 token that allows him to create a draft release. When you create the key as described in the tutorial I linked above, it will be encrypted with the private key for travis for this repository, so there is no problem of writing the key you get into the `.travis` file and commit them to this repository.
2. Benifits:
 * It simplifies the release process. A draft of the release we want to create is automatically created for us, containing the generated installers. We just have to fill in the changelog and hit the release button.
 * Less prone to add accidental dependencies to the executable when building and compiling the installer. The Travis server is a clean and controlled environment. It also doesn't change between releases.

gyunaev:
1. What if Travis private key leaks? Those things happen all the time.

2. I do builds on a dedicated VM which is builds-only and always launches from a snapshot. It is more controllable than Travis, actually.

Abestanis:
> What if Travis private key leaks? Those things happen all the time.

When was the last time that happend? I cant find anything on Google about it.

> I do builds on a dedicated VM which is builds-only and always launches from a snapshot. It is more controllable than Travis, actually.

It is not as comfortable, but if you don't accept this, I would prefer if you do, since you have such a  vm ready.

gyunaev:
This may have never happened (one wouldn't know), but the risk isn't worth the effort for me.
So either I build it, or if you can set Travis to upload them somewhere else.

Abestanis:
Another option is to replace Travis with Github Actions (in the future). I would be surprised if you would need to specify a key there to create draft releases. For now it is probably the easiest if you build it.

Abestanis:
Switching from Travis to Github Actions will take a bit, lets close this for now and create the installer manually.

PR #204, opened 2019-12-05
Update German translations

Abestanis:
Add the missing German translations.

I also found a string in the `MailAccountDialog` that was not translated.


gyunaev:
You can commit translation updates directly. Especially German ones :)

Issue #48, opened 2018-12-12
Ability to assign keyboard shortcut to show/hide


vlatkoB:
Hi, I have assigned a keyboard shortcut for launching TB, but when I launch it with that key, the "link" between Birdtray state of TB and the real state of TB seems to be lost, and the only way to "link" them again, so Birdtray can minimize and show TB again, is to click the icon couple of times.
After activating TB with the key, there are no two TBs active, only the same one.

It would be nice if we could bind a key to, say, `birdtray --show/hide` so Birdtray would do the same action as clicking on its icon, i.e. know what the current state of TB is and manage it.

Is something like that maybe already possible? What do you think?

gyunaev:
Note that you can hide Thunderbird without birdtray, i.e. by using xprop.

vlatkoB:
I just think it would be a nice thing to have it all in one tool, which already does that properly.

gyunaev:
Unfortunately global key binding (meaning the key is active even if the application is in background) is something which is system-dependent, and cannot be reliably implemented in Birdtray. I will provide show/hide options, and then you can bind them to keys using your OS-specific tool (such as Settings in KDE).

vlatkoB:
That's exactly what I meant. Thanks for looking at it.

6ber6:
I would be please (as it is the case with the extension minimizetotray that thunderbird minimizes also when we close it using the top right cross button, not only the minimize button. Because, currently, maany times I am using the cross button and thus birdtray as to start again Thunderbird. If somrone want to close definitively Thunderbird, he can either use the menu File->Exit or to configure birdtray to close Thunderbird when it exits. Thanks

gyunaev:
@6ber6 this is not possible.

Issue #96, opened 2019-06-15
Only show systray icon if there are new messages


goodwillcoding:
Thank you for birdtray. I was wondering if showing icon only if there are new messages is anywhere on the roadmap

gyunaev:
But since Birdtray itself has no window, how would you change this setting (or any other) if there is no email message?

Abestanis:
How about this:
If you start another birdtray instance, it will first check if there is an instance already running. If so, it would signal the running instance to open the settings dialog and the new instance will exit.

gyunaev:
Possible, but not very user-friendly. Let me think.

Abestanis:
Do we want to implement this now? The system to send a command to the running Birdtray instance is now in place.

gyunaev:
Yes, could be implemented now. Please show a dialogbox when this checkbox is set letting the user know that to remove that setting without receiving an email they have to start Birdtray again which will then show the settings window.

Abestanis:
Should we always show the settings, if another Birdtray instance is started while the first one is hidden? Or only if the second instance is started with the `--settings` command?

GerhardPehland:
Please make this a selectable option only. I am using the tray icon to open/close Thunderbird, and I want to be able to open Thunderbird, when there are no new messages too. To write a new email for one thing...

gyunaev:
Selectable and off by default. Probably in Advanced too.

PR #209, opened 2019-12-07
Added a way to pass arguments to running Thunderbird process

gyunaev:
This is used to implement toggle command support ( for #48 ). One unfortunate side effect of it is that if main instance of Birdtray is killed with -9, it will not restart until the shared segment is deleted (would require running it twice).

### DIFF
--- a/src/birdtrayapp.cpp
+++ b/src/birdtrayapp.cpp
@@ -133,6 +133,7 @@ bool BirdtrayApp::isInstanceRunning()
 
     autoUpdater = new AutoUpdater();
     trayIcon = new TrayIcon( mParser.isSet("settings"), sharedMemory );
+    return false;
 }
 
 BirdtrayApp* BirdtrayApp::get() {
--- a/src/birdtrayapp.cpp
+++ b/src/birdtrayapp.cpp
@@ -1,11 +1,20 @@
 #include <QtCore/QLibraryInfo>
 #include <QtCore/QStandardPaths>
+#include <QtCore/QSharedMemory>
+#include <QJsonObject>
+#include <QJsonDocument>
+#include <QMessageBox>
+
 #include "birdtrayapp.h"
 #ifdef Q_OS_WIN
 #  include "birdtrayeventfilter.h"
 #endif /* Q_OS_WIN */
 #include "utils.h"
 #include "morkparser.h"
+
+// Maximum memory size for inter-application communication
+static const int SHARED_MEMORY_SIZE = 4096;
+static const char * SHARED_MEMORY_TOKEN = "aeheewah7kahDauteeCeeP0jeipheiJuShailona0heidahyee2chahfiiLoh1ah";
 
 
 BirdtrayApp::BirdtrayApp(int &argc, char** argv) : QApplication(argc, argv) {
@@ -14,6 +23,7 @@ BirdtrayApp::BirdtrayApp(int &argc, char** argv) : QApplication(argc, argv) {
     QCoreApplication::setOrganizationDomain("ulduzsoft.com");
     QCoreApplication::setApplicationName("birdtray");
     QCoreApplication::setApplicationVersion(Utils::getBirdtrayVersion());
+
     // This prevents exiting the application when the dialogs are closed on Gnome/XFCE
     QApplication::setQuitOnLastWindowClosed(false);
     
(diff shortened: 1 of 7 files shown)


Abestanis:
I was researching on how to implement single instance support in Qt as well, and if found this blog post:
https://www.medo64.com/2019/07/single-instance-application-in-qt/

That implementation seems to be more robust. Also, using the `QLocalServer` would simplify communication between the two processes.

A few additional suggestions:
* Try to implement all the interprocess communication in the `BirdtrayApp` class, because that seems like the logical place to handle communication between two instances of the Birdtray app. Also, why did you move the call to `isInstanceRunning` into `main.cpp`, instead of calling it in the constructor of `BirdtrayApp`?
* Instead of just quitting when we detect another instance, how about we tell the other instance to show the settings. That would allow to implement #96. This should at least be done if the `settings` argument is specified.
* In addition to the `toggle` command, I think a `show` and `hide` command would be easy to add.

gyunaev:
I've seen this, but its too-much-hassle for a too-minor-issue. Especially with semaphores, which are horribly bad on Linux.

main.cpp needs to know when to exit, and constructors cannot return value.

We can add other commands later once the existing one works fine.

Abestanis:
There are no semaphores used on Linux, only on Windows.

I believe the implementation is worth it, because it would greatly simplify the interprocess communication. Instead of polling every so often, we can rely on the signals of the `QLocalServer`.

You can use `BirdtrayApp::exit(int code)` and a normal `return` to set the return code in the constructor.

gyunaev:
Signals from QLocalServer come from polling :) there is no some magic routine which would generate those signals otherwise.

You cannot return anything in the constructor, because it returns an object itself. And an app may decide to do something else before exiting.

Abestanis:
First of all, there are multiple ways to poll, some more efficient than others. For example, Qt already has a poll loop and would probably use select to wait for multiple events. The `QLocalServer` has access to these internals and could add the socket file descriptor to the list of watched file descriptors.

Secondly, the `QLocalServer` is probably calling [`accept`](http://man7.org/linux/man-pages/man2/accept.2.html), which to my knowledge is the most efficient way to wait on a socket, so it doesn't even have to poll.

Regarding exiting during the constructor, please look [here](https://github.com/gyunaev/birdtray/blob/master/src/birdtrayapp.cpp#L41) for an existing example of what I meant.

gyunaev:
If you instantiate QLocalServer, its Unix domain socket, and has to be polled via select/poll/epoll - I know of no other way to get information about when the socket is ready to accept a new connection. 

Having it stuck on accept() (I don't think it works like that, but anyway) would block the calling thread, meaning for this to work an app would start another thread(!) which would simply wait for a remote command - something 99% of users would never issue.

The example you provided is rather a hack (which is acceptable because the code is there for debugging and users wouldn't even know about it).

Abestanis:
I'm sorry, but why do you think it is a hack? What is wrong with it?

My point is that `QLocalServer` will select the best way to detect new messages for every operating system for us without us having to worry about it. Don't you think using `QLocalServer` will simplify IPC, especially if we want to add more commands which are not specific to the `TrayIcon` class?

Also, wouldn't it be nice, if we could drop the `mStateTimer` in the future and don't need to poll every one second?

Abestanis:
> One unfortunate side effect of it is that if main instance of Birdtray is killed with -9, it will not restart until the shared segment is deleted (would require running it twice).

The solution from the link handles the Birdtray crashing, whereas the current solution does not. Isn't that reason enough to switch?

gyunaev:
> I'm sorry, but why do you think it is a hack? What is wrong with it?

It doesn't work as expected. You expect `birdtray --decode aaa` to print decoded content and to exit the app, right? It doesn't; it hangs instead after printing decoded content.

Note that it worked fine in the original code (when it was in main) - its moving to QApplication subclass which broke it. Why? Because now it calls QApplication::exit, and it says:

```If the event loop is not running, this function does nothing```

and when the app starts, event loop doesn't yet run.

> My point is that `QLocalServer` will select the best way to detect new messages for every operating system for us without us having to worry about it.

There is no need to optimize that. We're not talking the main processing loop on a webserver here. This is functionality which maybe 1 out of 10,000 users would use maybe a few times an hour. Any optimization which goes into it is total waste of time.

> Also, wouldn't it be nice, if we could drop the `mStateTimer` in the future and don't need to poll every one second?

This is not going to happen, because we will not be able to accommodate for all possible events. For example there is no event you can sign for to see that Thunderbird is minimized or closed (unless you started it yourself, in which case you'll get QProcess signal).

**Comment by Abestanis** [comment]
> its moving to QApplication subclass which broke it

I swear I tested this and it worked fine. Sorry about that.

> There is no need to optimize that. [...] This is functionality which maybe 1 out of 10,000 users would use

Again, I'm sorry but this is simply not true. `checkSharedMemoryMessage` is called in `TrayIcon::updateIcon`, which is called by every Birdtray user at least one per second every second Thunderbird is running.

> there is no event you can sign for to see that Thunderbird is minimized or closed

This might be true for X11, but on Windows that is most certainly possible. We are using a callback based system right now.

But everything aside, the solution from the link handles Birdtray crashing. That seems to me like it is enough reason to prefer it over the current implementation.

**Comment by gyunaev** [comment]
> I swear I tested this and it worked fine. Sorry about that.

Its fine. As I said, its not worth fixing, this is debugging functionality. Invest your time where it matters.

> Again, I'm sorry but this is simply not true. `checkSharedMemoryMessage` is called in `TrayIcon::updateIcon`, which is called by every Birdtray user at least one per second every second Thunderbird is running.

Yeah, a few more CPU cycles used, not big deal. The most important thing is that it doesn't use much of my time. What I truly hate is receiving bugreports for this kind of functionality used by very few users, when it doesn't work for them. It feels like doing custom development.

> This might be true for X11, but on Windows that is most certainly possible. We are using a callback based system right now.

We have to play by lowest common denominator here.
 
> But everything aside, the solution from the link handles Birdtray crashing. That seems to me like it is enough reason to prefer it over the current implementation.

I don't think it does. Stale local socket files left in the filesystem is an issue, which requires hacks to work around for this kind of situation, where you have both the client and the server in a single package and unsure whether the user is running a client (i.e. second instance) or a server (first instance). It is only not an issue when you know you're starting the first instance.

**Comment by Abestanis** [comment]
> What I truly hate is receiving bugreports for this kind of functionality used by very few users, when it doesn't work for them.

I'd say the chances of receiving bug reports for this is equal no matter how we implement this.

> We have to play by lowest common denominator here.

No we don't. Why should we? We can easily extract the timer into `windowstools_x11.cpp`.

> Stale local socket files left in the filesystem is an issue

If I interpret the code correctly, `QLocalServer::removeServer` is used to handle exactly that situation.
Additionally, the code has the advantage that it handles calling multiple commands right after one another, e.g. ```bidtray command1 && birdtray command2```, where as the current implementation has a high chance of overwriting `command1` in the shared memory, before it is read by `checkSharedMemoryMessage`.

**Comment by gyunaev** [comment]
> I'd say the chances of receiving bug reports for this is equal no matter how we implement this.

No, because the code I used there works well in another project.

> No we don't. Why should we? We can easily extract the timer into `windowstools_x11.cpp`.

To keep the overall codebase simple. This is the main goal; being a free project I don't want to increase its complexity unless really needed. Was it commercial project, or University research project, things might have been different, but it is not.

> If I interpret the code correctly, `QLocalServer::removeServer` is used to handle exactly that situation.

Yeah, you can also call `unlink`. The problem is to decide whether to call this - you start a new app and the socket file is there. Does it mean there is an existing socket (and thus existing app)? Or this file has been left from an old app which was killed with SIGKILL and wasn't able to clean it up?

> Additionally, the code has the advantage that it handles calling multiple commands right after one another, e.g. `bidtray command1 && birdtray command2`

And why would this be used instead of `bidtray command1 command2`?

**Comment by Abestanis** [comment]
I tested your solution on Windows and it works nicely. It even works if Birdtray crashes, because in contrast to Unix systems on Windows the process does not own the `SharedMemory`, so it is automatically released if the process dies. The only thing that doesn't work is the `bidtray command1 && birdtray command2`. Some users will try that and will discover that it's broken.

Because of this and because it doesn't handle Birdtray crashing on Unix systems well, I implemented the solution from the blog post. By doing that, I realized I put too much faith into the writer of that block post, his solution only works on Windows.
But with this little addition:
```c++
...
if (!serverListening && (server->serverError() == QAbstractSocket::AddressInUseError)) {
    QLocalSocket serverSocket;
    serverSocket.connectToServer(SINGLE_INSTANCE_SERVER_NAME, QLocalSocket::WriteOnly);
    if (serverSocket.waitForConnected()) { // There is a default timeout of 3 seconds, which we can change
        sendCommandsToSingletonApp(serverSocket);
        return;
    }
    QLocalServer::removeServer(SINGLE_INSTANCE_SERVER_NAME); //cleanup
    serverListening = server->listen(SINGLE_INSTANCE_SERVER_NAME); //try again
}
...
```
it seems to work really well, solving both of the above problems on Windows as well as on Unix systems.

If you want to check it out, it's the [feature/single_instance](https://github.com/Abestanis/birdtray/tree/feature/single_instance) branch. Please tell me what you think. The code complexity seems to be roughly the same to the current approach, but I thought it be easier to keep sending and receiving of the IPC commands both in `BirdtrayApp`.

gyunaev:
Ok, so the workaround here is to wait for 3 seconds. If UI thread is blocked for longer (which may happen), it will fail. Unsure what advantages this approach brings, if any.

Abestanis:
Thank you for your time in testing this. However, I could not replicate what you found.

First of all, the solution is not to wait for three seconds. In my testings, I wasn't able to make it wait at all, even if the UI thread was deadlocked. Qt seams to use a different thread for the server.

Since it looks like I didn't do a great job at explaining the advantages of the other approach over the current approach, let's do a comparison:

| Scenario | Current approach | New approach |
|----------|------------------|--------------|
| Birdtray crashes. The user starts a new instance. | Windows: The new instance will start normally. <br> Unix: The new instance will exit, as if there were still another instance running. | All platforms: The new instance starts normally. |
| Birdtray freezes. The user starts a new instance. | The new instance will exit. | The new instance will exit. |
| Birdtray freezes for a short time, then unfreezes. During the freeze time, the user starts two instance of Birdtray with `command1` and `command2`. | When Birdtray unfreezes, only `command2` is executed, `command1` was overwritten. | When Birdtray unfreezes, both commands are executed. |
| The user starts two instances of Birdtray within a second, e.g `birdtray command1 && birdtray command2`. | There is a high chance that `command1` is not executed, because it is overwritten by `command2`. There is a small change that `command1` is read before it is overwritten by `command2`. The behaviour is *non deterministic*. | Both commands will be executed. The behaviour is *deterministic*. |
| Performance | To check the for new commands, we require a timer that fires every one second. | Qt chose an optimized strategy for every operating system. We don't need to concern us with the specifics, we let Qt handle it. But I am confident that they have a better way than polling every second. |
| Code complexity | Sending of commands is handled in `BirdtrayApp`, receiving of commands is handled in `TrayIcon`. The separation of both makes it easier for them to diverge. | Both sending and receiving of commands are handled in `BirdtrayApp`. |

I hope this gives you a bit of insight into my thought process on why I prefer the new approach.

pablo-blueoakdb:
With my new `gen_birdtray_accounts` shell script, it would be great if I could signal `birdtray` to flush its .config file, my script manipulates the .config file and after signals `birdtray` to re-open it.  

For now I'm asking the user to shutdown `birdtray` before running my script.

gyunaev:
Can you please create a PR with your change? If you have tested it on both platforms, we can use yours.

PR #210, opened 2019-12-10
Elaborate confugration file location in README

Chadsr:
This PR simply elaborates in the README as to where the Birdtray configuration file is saved, to avoid confusion since the naming is unexpected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,17 @@ You can also enable birdtray to start Thunderbird when you start Birdtray, or en
 
 Once you change settings, often you need to restart birdtray for the new settings to take effect.
 
-All birdtray settings are stored on a per-user basis in `$HOME/.config/ulduzsoft/birdtray.conf`.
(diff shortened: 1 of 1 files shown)


Abestanis:
Thank you for your contribution. Could you please clarify, that this is the default path for Linux? On Windows, it's in the registry at `HKEY_CURRENT_USER\Software\ulduzsoft\birdtray`. @ThanosApostolou  Is the config path the same if Birdtray is installed via Flatpak?

ThanosApostolou:
When installed as flatpak, the configuration is at `$HOME/.var/app/com.ulduzsoft.Birdtray/config/ulduzsoft/birdtray.conf`

Chadsr:
I have added Flatpak and Windows locations to the PR. Let me know what you think.

Abestanis:
Looks good to me. 👍

PR #211, opened 2019-12-10
Show warnings about problems parsing mail accounts as a tooltip on the tray icon

Abestanis:
Previously, we showed a red `X` in the icon, if the parsing of one account failed, even if we could parse the unread count from other accounts.

We now don't abort the parsing process, if only one account couldn't be parsed. If a problem occurs, we show an exclamation mark in the icon:
Icon with warning
Icon with unread count and warning
Additionally, a tooltip is added with the problem description:
Warning tooltip
<details>
<summary>Problematic accounts are also marked in the settings.</summary>

Accounts list with warning
</details>

If anybody has a better color combination for the exclamation mark, let me know. I'm not an artist and this is the best I could come up with that was still readable on the small icon.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@ endif(NOT Qt5LinguistTools_FOUND)
 if(WIN32)
     set(EXECUTABLE_OPTIONS WIN32)
     ENABLE_LANGUAGE(RC)
-    add_definitions(-DUNICODE)
+    add_definitions(-DUNICODE -DNOMINMAX)
     set(PLATFORM_SOURCES src/windowtools_win.cpp src/birdtrayeventfilter.cpp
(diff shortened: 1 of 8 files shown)


gyunaev:
Can you just append "!" to the unread email count? This way it would be drawn as a text line and always visible (because the users would choose contrast font color and icon). Would be a bit simpler too.

Abestanis:
That is a good idea. I tried it out, but I'm a bit worried about the visibility.

I don't mean the contrast to the background. This is supposed to be a warning indicator, but it is not really recognizable if it is the same color as the unread counter text. It also gets really hard to notice, once the number of unread mails is greater than 99. I've made a few screenshots, the first four were made with the current code, the other ones with your suggestion.


1Current 10Current 100Current 1000Current

1New 10New 100New 1000New

Issue #191, opened 2019-11-28
Feature request: Option to always start ignoring unread mails


Mintulix:
**Description**
I'd like to see the behaviour to only show new mails during a session and ignore old unread mails without selecting this manually from the menu.
There should be an option or command-line start parameter to ignore these unread mails automatically.

**Example**
44 unread mails in your selected msf-accounts when starting birdtray.
During a session two new mails are coming in -> counter should count "2" and/or icon is changing to show there are new unread mails.
When I terminate birdtray and start it again, the counter should be at zero again. Also zero when I reduce the unread mails "below zero".

gyunaev:
You're welcome to submit a patch.

Mintulix:
Ok, I think I found a solution, at least for me (without GUI option to turn it off):

In trayicon you need two new variables, initialized like that:
```
TrayIcon::TrayIcon(bool showSettings)
...
    mFirstRun = true;
    mMyUnreadEmails = 0;
```
Code changes are in unreadCounterUpdate only:
```
void TrayIcon::unreadCounterUpdate( unsigned int total, QColor color )
{
    // total at this point and in every loop is the real actual unread email value (e.g. 26)
    // It is manipulated here to set the tray value mUnreadCounter to another value (zero at the beginning)
    if (mFirstRun) {
        // Saving the old value (just on the first loop)
        mMyUnreadEmails = total;
        // Set unread emails to zero
        total = 0;
        mFirstRun = false;
    }
    // If the new value is greater then the saved one, show the difference (e.g. 2 = 28 - 26)
    if (total > mMyUnreadEmails) {
        total = total - mMyUnreadEmails;
    } else {
    // If the new value is equal or less then the saved one, leave it at zero and adjust mMyUnreadEmails to the new value
    // (e.g. read 4 of my 28 unread emails -> new base value is 24. Tray value is zero. New email comes in -> 1 = 25 - 24)
        if (total > 0) {
            mMyUnreadEmails = total;
            total = 0;
        }
    }

    Utils::debug("unreadCounterUpdate %d", total );
    mUnreadCounter = total;
    mUnreadColor = color;

    updateIcon();
}
```
For other users the new code-part has to be surrounded by an option that could be switched in the GUI.

gyunaev:
As I have said, please submit a proper patch via pull request.

Abestanis:
@Mintulix If you don't feel like you are up to the task, I could create a pull request for you, when I have the time.

[...]

PR #212, opened 2019-12-11
Add option to ignore unread mails at startup (by @Mintulix)

Abestanis:
This implements @Mintulix idea from #191 to add an option to ignore unread emails that already exist when Birdtray starts.

This is disabled by default. It can be enabled via a settings in the monitoring tab:
<details>
<summary>Image</summary>

Monitoring tab in settings with new option
</details>

@Mintulix Can you please test that this still works as you intended?


Mintulix:
Hi @Abestanis , that was quick! I made the same tests like with my code, and it seems to work great. Thank you also for refining the code.
I hope someone else finds this option useful...

PR #213, opened 2019-12-18
For issue 126, use this script

pablo-blueoakdb:
This _pull request_ requests two items:

1. The creation of **contrib/** and
2. Seeding it with the new shell script _gen_birdtray_accounts_

ps:  This is my first ever use of github so all feedback welcomed!


gyunaev:
Please create a subdirectory, i.e. contrib/auto-config-generator or something like that, and move the files there. You can commit into the same pull request.

PR #222, opened 2020-01-07
Add a "Don't ask again" button to the "No accounts configured" dialog

### LINKED ISSUE #216: make birdtray usable without using its monitoring capabilities
If you do not configure monitoring (or had it configured but chose to "disable" it again), birdtray keeps asking if you'd like to set up folders for monitoring on startup. Due to a few open issues, I'd like to resort to Thunderbird's notifications entirely and leave monitoring off in birdtray.

If this behaviour is by design, I request to change it, for example ask only once and refer to the settings. If its a bug, you now know about it. :)

Abestanis:
This closes #216. The new `Don't ask again` button allows to disable the dialog.

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -46,6 +46,7 @@ Settings::Settings(bool verboseOutput)
     mLaunchThunderbirdDelay = 0;
     mShowUnreadEmailCount = true;
     ignoreStartUnreadCount = false;
+    showDialogIfNoAccountsConfigured = true;
     mThunderbirdWindowMatch = "- Mozilla Thunderbird";
     mNotificationMinimumFontSize = 4;
     mNotificationMaximumFontSize = 512;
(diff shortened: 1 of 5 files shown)


gyunaev:
The checking code is kinda ugly. Maybe better to add:

questionDialog.addButton( tr("Don't ask again"),  QMessageBox::DestructiveRole );

and then check questionDialog.result() to see which one of three buttons was clicked? Would probably streamline/simplify the code a bit.

Abestanis:
That could work, but [`MessageDialog::result()`](https://doc.qt.io/qt-5/qdialog.html#result) returns ` QMessageBox::StandardButton::NoButton` if the user presses the `Don't try again` button, which would make the check confusing and also doesn't change mutch of the checking code:

```c++
questionDialog.addButton(tr("Don't ask again"), QMessageBox::DestructiveRole);
showSettings = questionDialog.exec() == QMessageBox::Yes;
if (questionDialog.result() == QMessageBox::NoButton) {
    settings->showDialogIfNoAccountsConfigured = false;
    settings->save();
}
```

gyunaev:
switch ( questionDialog.result() )

 case QMessageBox::Yes:
   show settings

case QMessageBox::NoButton:
  set the flag and save

Abestanis:
Ah, you want to completely replace the `No` button with the `Don't try again` button?

gyunaev:
No, "no" would do nothing in this switch as there's no action.

Abestanis:
Sorry, I don't see how you would integrate the other checks into the switch. I think you should be able to push changes to this branch if you want to make the changes yourself, otherwise it's probably easier if you close this pull request and you create a new one.

gyunaev:
Will change it later.

PR #225, opened 2020-01-07
Add missing German translation 🌐

Abestanis:
This adds a missing German translation that was supposed to be included in #223.


gyunaev:
Please commit translations and documentation updates directly :) no need for pull req for this.

Abestanis:
Ok, will do so in the future. :)

PR #228, opened 2020-01-10
Clarify the show, hide and toggle command line options

Abestanis:
I propose to add a `-thunderbird` suffix to the name of the three command line options, to prevent confusion about their meaning. This shouldn't really affect anybody, since these commands were added only two days ago.

`--toggle` → `--toggle-thunderbird`
`--show` → `--show-thunderbird`
`--hide` → `--hide-thunderbird`

### DIFF
--- a/src/birdtrayapp.cpp
+++ b/src/birdtrayapp.cpp
@@ -9,9 +9,9 @@
 #include "windowtools.h"
 
 #define SINGLE_INSTANCE_SERVER_NAME "birdtray.ulduzsoft.single.instance.server.socket"
-#define TOGGLE_THUNDERBIRD_COMMAND "toggle-thunderbird"
-#define SHOW_THUNDERBIRD_COMMAND "show-thunderbird"
-#define HIDE_THUNDERBIRD_COMMAND "hide-thunderbird"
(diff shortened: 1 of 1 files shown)


gyunaev:
Please no:

- Changing command-line options after they are published is bad practice;
- Making them longer makes no sense as there is no potential for confusion.

Abestanis:
I agree that renaming command line options is generally a bad idea, but in this case we didn't really publish a release. As I mentioned, they were only merged into master two days ago.

I believe there is a potential for confusing. Without reading the help message it is unclear what the `toggle` command toggles. It will also be very confusing if we add another command line option to toggle something else within Birdtray.
The same goes for the other two. If we implement #96, people might think they need to run `birdtray --show` to make the tray icon visible again.

gyunaev:
As Linux user one thing I really dislike is to type really long command-lines when unnecessary. 
rm --remove-directories-too --force-removing-them-if-not-empty - no thanks.

Abestanis:
Would you be ok with `show-tb` etc.?

gyunaev:
I'd be even better with -s, -H and -t :)

Abestanis:
I can add these too, give me a sec.

PR #233, opened 2020-01-21
Add setting to hide the Birdtray icon if no new mail is present

### LINKED ISSUE #96: Only show systray icon if there are new messages

Abestanis:
Resolves #96. This adds a new option in the advanced tab, which is disabled by default.
If enabled, the Birdtray system icon is hidden, if no new mails are detected. It will also be hidden if the user uses the snooze feature, is that ok with you @gyunaev?

If the Birdtray icon is hidden, one can access the setting to turn off the option by starting a second instance of Birdtray with the `--settings` parameter. Note that just starting Birdtray without a parameter does nothing.

<details>
<summary>New setting in the advanced tab</summary>

Advanced settings tab
</details>


Abestanis:
Looks like Qt has changed their online installer again, so the Windows builds time out.

PR #247, opened 2020-02-08
Attach to the parent console if the --debug parameter is present on Windows

### LINKED ISSUE #244: BirdTray - Log File to show unread emails
Thunderbird : 68.3.0
O/S : Windows 7 64 bit
BirdTray : 1.7.0

BirdTray does not update the Email Counter once all the emails are read.  How do I enable logging to see which mail folder it thinks that there are unread emails ?

Abestanis:
Previously, it was not possible to view the debug output of Birdtray on Windows without attaching to the process with a debugger. On Windows, a program can either be compiled as a GUI application or a command line application. A command line application will automatically send its output to an attached console, but will also automatically create a console if it is started without one, meaning that Birdtray would open a cmd window if it was started. Because of this Birdtray is compiled as a GUI application, which doesn't create a console, but also doesn't automatically attach to one.

We now attach to the parent console, if there is one and if the `--debug` parameter was specified.
This means that the output is now visible on the console (closes #244), but unfortunately the command line still starts Birdtray in a detached mode and doesn't wait for any output. That means that the Birdtray output will appear after the new prompt. To avoid that, one can launch Birdtray using the [start command](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/start):
```cmd
start /w birdtray.exe --debug <arguments>
```
terminal

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -10,11 +10,13 @@ int main(int argc, char *argv[]) {
 #ifdef Q_OS_WIN
     for (int i = 1; i < argc; i++) {
         if (strcmp("--debug", argv[i]) == 0) {
-            AttachConsole(ATTACH_PARENT_PROCESS);
-            // reopen the std I/O streams to redirect I/O to the parents console.
-            freopen("CON", "w", stdout);
-            freopen("CON", "w", stderr);
-            freopen("CON", "r", stdin);
+            if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+                // reopen the std I/O streams to redirect I/O to the parents console.
+                FILE* newFile = nullptr;
+                freopen_s(&newFile, "CON", "w", stdout);
+                freopen_s(&newFile, "CON", "w", stderr);
+                freopen_s(&newFile, "CON", "r", stdin);
+            }
             break;
         }
     }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,8 +1,24 @@
 #include <QApplication>
 #include "birdtrayapp.h"
-
+#ifdef Q_OS_WIN
+#  include <windows.h>
+#  include <cstdio>
+#  include <cstring>
+#endif /* Q_OS_WIN  */
 
 int main(int argc, char *argv[]) {
+#ifdef Q_OS_WIN
+    for (int i = 1; i < argc; i++) {
+        if (strcmp("--debug", argv[i]) == 0) {
+            AttachConsole(ATTACH_PARENT_PROCESS);
+            // reopen the std I/O streams to redirect I/O to the parents console.
+            freopen("CON", "w", stdout);
+            freopen("CON", "w", stderr);
+            freopen("CON", "r", stdin);
+            break;
+        }
+    }
+#endif /* Q_OS_WIN */
     BirdtrayApp app(argc, argv);
     return QApplication::exec();
 }


gyunaev:
I was thinking to simply pop up a QTextBrowser with debug log output. Do we really need the console?

Abestanis:
Sounds more complicated than this for me, but your call.

gyunaev:
Easier for new use to copy-paste logs, and same UI on all platforms. Seems better.

Abestanis:
Ok, I'll leave it to you.

Abestanis:
Well, on second thought, how will a user access the log if birdtray crashes or never shows a UI?

Abestanis:
@gyunaev Should we reopen this or did you find a solution to these problems?

Abestanis:
I think, until we come up with a better solution, this is our best bet.

gyunaev:
I'm still thinking about this one.

gyunaev:
Added a7fff0b

Abestanis:
Nice, I really like the new logging system. It is really good from a users perspective.
It doesn't however solve that problem that the user can't provide us any logs if Birdtray crashes (not via `Log::fatal`).
From a Windows developers perspective, I would also like to see the output in my terminal or IDE window while debugging, but at this point it is just a minor inconvenience. Do you have any reservations against merging this in addition to the new log system?

gyunaev:
Would adding a call to OutputDebugString be enough to have it in your IDE?

From my past memories Windows terminal logging is very awful to use, but if you want it, you can add it.

gyunaev:
The advantage of this system (versus writing a log file) is also you can now dump all kind of messaging there, and they won't fill up anything.

Abestanis:
> Would adding a call to OutputDebugString be enough to have it in your IDE?

Unfortunately not, the log would only get captured if I run Birdtray with a debugger attached, which I don't want to do all the time. You know how slow programs get with a debugger attached.

> The advantage of this system (versus writing a log file)

Yes, that is one of the reasons I like your approach.

Abestanis:
By the way did you see, we closed **9** issues this week? 😄🎉 
10 if you count #244.

gyunaev:
Yeah. We're all locked up in houses here in California. How's Germany?

gyunaev:
> > Would adding a call to OutputDebugString be enough to have it in your IDE?
> 
> Unfortunately not, the log would only get captured if I run Birdtray with a debugger attached, which I don't want to do all the time. You know how slow programs get with a debugger attached.

What about https://docs.microsoft.com/en-us/sysinternals/downloads/debugview ?

In any case, instead of --debug popping up console, I think we might benefit from --debug actually writing a log into a file, what do you think?

Abestanis:
> Yeah. We're all locked up in houses here in California. How's Germany (I assume you're there)?

We're not locked in (yet), but we are all advised to stay at home as much as possible. Luckily I can work from home.

> What about https://docs.microsoft.com/en-us/sysinternals/downloads/debugview?

That could be an option, but I don't see the advantage over the terminal.

> In any case, instead of --debug popping up console

`AttachConsole` does not create a console, it only connects to one if one was already present (e.g. Birdtray was started from the command line). It will not pop up a new console.

> I think we might benefit from --debug actually writing a log into a file, what do you think?

That is a good idea, if a user starts Bidtray with the `--debug` option, they probably want to get the log.

gyunaev:
Same here, also working from home.

Ok, let's do the --debug to attach and --log to write log?

Abestanis:
Yeah, I think splitting it into two separate arguments is a good idea. Do you want to work on logging to the log file? Since you wrote the logging code recently, I think it will be easier for you to integrate it there. Do you need me to do anything for this pull request?

gyunaev:
I can add logging indeed. Shall this be merged?

Abestanis:
Yes please. 😄

PR #254, opened 2020-02-19
Make tray icon size consistent

dvalter:
Fix inconsistent drawing of unread counter across different sizes of icons by
scaling icons to a predefined size.

In case of small icons (24x24, 32x32), unread count and border may be rendered incorrectly.
In addition, default icon size was used for the custom unread icon, so it caused image to be either too small or too big.

This solution scales an icon to a fixed size (original 128x128 is used) before drawing it. It may be an overhead, but it should work with existing configurations.

However I have a different solution [here](https://github.com/dvalter/birdtray/tree/icon-scaling-fix-v2), which scales icons before applying them to Settings from DialogSettings. The only downside I see is requirement for user to re-apply settings for changes to take effect.

### DIFF
--- a/src/dialogsettings.cpp
+++ b/src/dialogsettings.cpp
@@ -491,7 +491,12 @@ void DialogSettings::changeIcon(QToolButton *button)
         return;
     }
 
-    button->setIcon( test );
+    Settings* settings = BirdtrayApp::get()->getSettings();
+
+    // Force scale icon to the expected size
+    button->setIcon( test.scaled(
+            settings->mIconSize.width(), settings->mIconSize.height(),
+            Qt::KeepAspectRatio, Qt::SmoothTransformation) );
 }
 
 
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -29,7 +29,7 @@ Settings::Settings(bool verboseOutput)
 #endif
 
     mVerboseOutput = verboseOutput;
-    mIconSize = QSize( 128, 128 );
+    mIconSize = QSize( ICON_SIZE, ICON_SIZE );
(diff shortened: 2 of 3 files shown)


Abestanis:
This is very good idea. 👍 

I am personally in favour of the second approach. I dislike the overhead every time the icon is updated and I think not changing anything for existing setups is a big bonus. Users might get confused if their icon suddenly looks different after the update.

Regarding the failed build, there is a problem with the action that installs Qt on the build server. If you rebase on master, you will get the workaround for that.

dvalter:
I rebased and pushed that second version, so builds look fine now.

> Users might get confused if their icon suddenly looks different after the update.

I should point out that in this case users still may experience a suggen change if they apply settings without touching an icon. If scaling code would be added to `DialogSettings::changeIcon` instead, this isn't a case, but referencing `Settings` from that method looks somewhat ugly for me.

Abestanis:
> If scaling code would be added to DialogSettings::changeIcon instead, this isn't a case, but referencing Settings from that method looks somewhat ugly for me.

What do you mean by that? I think, the following would work:
```c++
    Settings* settings = BirdtrayApp::get()->getSettings();
    button->setIcon(test.scaled(settings->mIconSize.width(), settings->mIconSize.height(),
            Qt::KeepAspectRatio, Qt::SmoothTransformation));
```

If you don't like the look of that, you could make `Settings::mIconSize` a static const member, it isn't changeable anyway at the moment.

dvalter:
I was worried about this from an architectural standpoint. Since it's totally fine, that's even better in terms of UX.

PR #263, opened 2020-02-22
Stop showing database access error when Mork is already 

dvalter:
Change signal-slot connection order to prevent from loading accounts list before parser is configured on start of the settings window.


Abestanis:
This is a good fix 👍. To prevent similar issues, we should also move the whole `// Parsers` block above the `// Setup parameters` block, so `isMorkParserSelected` returns the correct value from the start.

PR #266, opened 2020-02-26
spec file for rpmbuild

raven-kg:


gyunaev:
Technically we do not require Thunderbird; I'm unsure if this is needed.

rnc:
How about using `recommends` instead ? i.e. use a weak dependency ?
See https://rpm.org/user_doc/dependencies.html and https://fedoraproject.org/wiki/Packaging:WeakDependencies

raven-kg:
> How about using `Recommends` instead ? i.e. use a weak dependency ?
> See https://rpm.org/user_doc/dependencies.html and https://fedoraproject.org/wiki/Packaging:WeakDependencies

Recommends works only for newer versions of Fedora and CentOS 8. Building for older versions will fail with an error. And finally I agree with @gyunaev - there may be no Thunderbird installed from repos.

PR #274, opened 2020-03-18
Fix duplicated watched mail folders

### LINKED ISSUE #231: Account monitoring always  loses 5th account for list of accounts to monitor.
**OS**:  Ubuntu Mate 18.04
**Birdtray version**: 1.7.0
**Thunderbird version**: 68.2.2
**Birdtray origin**: From Source
**Qt version**: (sorry, not sure how to find this)

**Description**
I have 5 mailboxes in Thunderbird. I have added each of the Inbox folders to the "Monitoring" tab in birdtray. However whenever I reboot the last one in the list has always gone and I have to re-add it.

**Expected behavior**
All 5 inboxes should still be in the monitoring list.

**To Reproduce**
1. Add 5 mailbox inboxes to monitor
2. Reboot, only 4 will be in the list.

Abestanis:
This removes the ability to watch the same mail folder multiple times. Fixes #231.
This bug was possible, because we store the information about the number of watched folders twice in the settings; Once as the actual data (`accounts/accountX[Color|URI]`) and once as a number (`accounts/count`).
The duplication of information is what allowed this bug, which is why this pull request removes the `accounts/count` value, so the actual data is the single source of truth. However, we still read the `accounts/count` for old installations, to keep backwards compatibility.

### DIFF
--- a/src/modelaccounttree.cpp
+++ b/src/modelaccounttree.cpp
@@ -110,6 +110,11 @@ void ModelAccountTree::addAccount(const QString &uri, const QColor &color)
     if (uri.isEmpty()) {
         return;
     }
+    int existingIndex = mAccounts.indexOf(uri);
+    if (existingIndex != -1) {
+        editAccount(createIndex(existingIndex, 0), uri, color);
+        return;
+    }
     // Only this line changed
     beginInsertRows( QModelIndex(), mColors.size(), mColors.size() + 1 );
 


gyunaev:
Someone can still add the same account twice via symlinks or relative paths (i.e. c:\users\tb\account1 and c:\users\tb\..\tb\account1) - is this a concern?

Abestanis:
I don't think so, it is highly unlikely that someone will do that by accident. In any case, Birdtray will still work with even with the duplicated entries if they have different paths and will not delete other valid entries as seen in #231, which is the main goal of this pull request.

gyunaev:
Sounds good then. I'm unsure why you removed accounts/count and instead relying on iterating until the first empty element. Is it fixing some bug?

Abestanis:
I want to remove the duplicate information to prevent bugs like this in the future.

gyunaev:
But why the old code wasn't good for this? As long as your accounts are in order, the settings kept them in order. I.e. even if you had entries like accounts/URL5 they are ignored if your accounts/count is 3. I just don't understand why this code in settings is needed at all.

gyunaev:
I mean, if you prevent duplicate entries in edit and remove them on load, why bother removing them from settings storage too? This only makes sense if you're trying to save the space for the configuration file, but this is not worth bothering.

Abestanis:
Well, the old code didn't work correctly, hence the bug report #231. My goal is to prevent a bug like this to ocour in the future. I believe by removing the `count` we reduce the chance for mistakes.
I also liked the side effect that it cleans the config file. Not for the space, but it can be pretty confusing for a user if they would open the config file and find all their "deleted" old monitoring entries still in the file. At least I was confused for a second when I saw them. Obviously a typical user will never interact with the config file, but it is still a nice to have.

gyunaev:
Personally I'd prefer to have count and not remove unused entries (they indeed could be confusing for someone unfamiliar with source code, but they are not supposed to be user-editable anyway).

Abestanis:
Ok, I will remove the change. I just wanted to future-proof the settings, but I guess we can also just be careful to not repeat a mistake like this again.

gyunaev:
Actually in my other Github project Spivak I moved to have all settings stored in a JSON file. If we do this, we can get rid of count variables, since JSON has proper support for arrays.

Abestanis:
> ... have all settings stored in a JSON file

That sounds like a good idea, if it doesn't introduce a huge dependency. 👍

gyunaev:
It won't, JSON manipulation is already part of Qt.

Issue #220, opened 2020-01-01
Birdtray-1.7.0-Win-x64 faild to start on win10-x64


Harreth:
After the installation of Birdtray-1.7.0-Win-x64.exe
win10 x64 Version 1909 (OS Build 18363.535)
Thunderbird V68.3.1 X64 installed on system
launching the installed birdtray app results to this error:
(birdtray.exe - System Error
The code execution cannot proceed because 
VCRUNTIME140_1.dll was not found. Reinstalling the program 
may fix this problem.)
I Tryed to re install for all users, current user, run with compatibility, ended up with the same error message. 

thanks

MarkFirewhal:
This was closed before an answer was reported:
https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads

You should be able to fix this problem by downloading them directly from Microsoft.

PR #281, opened 2020-03-22
Detect missing C++ Runtime on Windows

Abestanis:
On a freshly installed Windows machine with no other programs, there will not be a C++ runtime and Birdtray will fail to start. The C++ runtime is shared by all programs on a Windows PC. If you install any other program (e.g. `Chrome` or `Firefox`), most of them will install the C++ runtime. That is why this is not a big problem and almost nobody encountered it, but it caused #220.

The installer for `1.7.0` contained an installer for the C++ runtime, but never installed it. Because the runtime installer made up a significant portion of the size of the Birdtray installer (~10Mb, 50% of the installer) and it is never used by 99.9% of the people using the installer, I decided the best option is to not include the C++ runtime installer in the Birdtray installer. Instead display a message and open the link to download the C++ runtime installer from the Microsoft servers. This will keep the installer size down and does not automatically downloads and runs an executable from the internet.

<details>
<summary>Images</summary>

"C++ runtime not found"-dialog

If the user clicks `No`, they get the following warning. In that case they can install the runtime manually:
"Birdtray not runnable warning"-dialog

If the user clicks `Yes`, the direct url to download the C++ runtime installer is opened in the default browser and the following message is displayed:
"Waiting for installation"-dialog

If the C++ runtime is still not detected after that, the user is represented with the following dialog:
"C++ runtime still not found"-dialog

In that case and if the user decided not to download the C++ runtime installer before, they don't get the option to start Birdtray from the installer:
"Installation finished"-page without the option to start Birdtray

Otherwise, they get a dialog that the C++ runtime was successfully detected.
C++ runtime detected dialog
</details>

### DIFF
--- a/installer/Utils.nsh
+++ b/installer/Utils.nsh
@@ -212,6 +212,45 @@ FunctionEnd
 !define SelectSection '!insertmacro SelectSection'
 !define UnselectSection '!insertmacro UnselectSection'
 
+!ifndef UNINSTALL_BUILDER
+
+# Check if a Visual C++ Redistributable is installed.
+# $0 - ARCH: The path of the directory (x64 or x86).
+# Returns:
+# $0 - IS_INSTALLED: 1 (true) if a Visual C++ Redistributable is installed and 0 (false) if not.
+Function IsVisualRedistributableInstalled
+    Exch $0
+    ${if} ${RunningX64}
+        ReadRegDword $0 HKLM "SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\$0" \
+            "Installed"
+    ${else}
+        ReadRegDword $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\$0" "Installed"
+    ${endif}
+    Exch $0
+FunctionEnd
+!macro IsVisualRedistributableInstalled VAR ARCH
+    Push ${ARCH}
+    Call IsVisualRedistributableInstalled
+    Pop ${VAR}
+!macroend
+!define IsVisualRedistributableInstalled "!insertmacro IsVisualRedistributableInstalled"
+
+
+# Open an url in a browser window.
+# $0 - URL: The url to open.
+Function OpenURL
+    Exch $0
+    Exec "rundll32 url.dll,FileProtocolHandler $0"
+    DetailPrint "$(OpenUrl)"
+    Pop $0
+FunctionEnd
+!macro OpenURL URL
+    Push ${URL}
+    Call OpenURL
+!macroend
+!define OpenURL "!insertmacro OpenURL"
+!endif # UNINSTALL_BUILDER
+
 # === Macros === #
 
 # Check that the program is suitable for the platform.
--- a/installer/Utils.nsh
+++ b/installer/Utils.nsh
@@ -215,7 +215,7 @@ FunctionEnd
 !ifndef UNINSTALL_BUILDER
 
 # Check if a Visual C++ Redistributable is installed.
-# $0 - ARCH: The path of the directory (x64 or x86).
+# $0 - ARCH: The architecture of the C++ Redistributable (x64 or x86).
 # Returns:
 # $0 - IS_INSTALLED: 1 (true) if a Visual C++ Redistributable is installed and 0 (false) if not.
 Function IsVisualRedistributableInstalled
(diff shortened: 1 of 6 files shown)


gyunaev:
Unfortunately the people who would need to install C++ runtime would be likely the people unlikely to be able to install it separately.
I suggest releasing two packages:
- The install package with C++ runtime included;
- The "advanced/portable" package, which is just a ZIP archive of install dir and without C++ runtime.

Abestanis:
> Unfortunately the people who would need to install C++ runtime would be likely the people unlikely to be able to install it separately.

In reality, this is not the case. Almost nobody will ever encounter this, only people who just installed Windows on their PC and who chose Birdtray as the first program to install.

> ... to be able to install it separately.

They also don't need to install it separately, they click yes in the dialog and the default browser pops up (since they didn't install `Chrome` or `Firefox`, that would be the `Edge` browser) asking them if they want to run the downloaded installer. They only have to click `Run` and the C++ runtime installer opens. They should be able to do that.

> I suggest releasing two packages:

I really want to avoid that, because it can cause confusion on what version the user needs.

gyunaev:
> I really want to avoid that, because it can cause confusion on what version the user needs.

ZIPs are also useful as portable versions; I'd like to have them regardless.

Abestanis:
Why do you need a portable version? Isn't it enough to not need admin rights during installation?

gyunaev:
This allows people who do not have their own assigned computer to put it together with Thunderbird on a USB stick, and carry it around. Then when they want to work on their emails, they plug it in, run both, and use it. This is common in poorer countries where people don't have personal computers.

Abestanis:
Yes, but you can already do that with the installer: Choose the `Current user only`-mode and change the installation directory to the USB stick.

jejo86:
I just installed `Birdtray-1.8.0-Win-x64.exe` on a PC running Win7, which is in use for at least two years now since the last fresh install and with several applications installed, among others _Firefox v75_. 

During the installation of Birdtray I was not warned about a missing C++ runtime as the latest release claims to do be doing from now on. Instead I got the typical `VCRUNTIME140_1.dll` not found error after starting the application.

Maybe you would like to review the effectivity of the change:
> The Windows installer now checks if the VS C++ Runtime is installed. Contributed by @Abestanis.

Abestanis:
Hm, thank you for reporting this. I don't have a Windows 7 PC so I wasn't able to test this on that configuration. Could you please open a new issue and provide the following details:
* The contents of the `HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\x64`
* Whether you do have any version of the `Microsoft  Visual C++ Redistributable` installed on you System. Check the `Programs and Features` page in the `Control Panel`. Search for entries like these:
"Microsoft  Visual C++ Redistributable" entries

jejo86:
I just installed the latest redistributables so I don't know whether the information is still of use to you: [BirdtrayIssue281.zip](https://github.com/gyunaev/birdtray/files/4532849/BirdtrayIssue281.zip)

On a fresh Windows one of the first things I do is to install the latest VC redistributables since many applications complain about missing dependencies. 

Downloading the latest redistributables I noticed it is a combined version for the years 2015 to 2019. I was pretty sure to already have had installed the 2015 Update 2 version, but maybe one of the first releases... The 2015 Update 2 files were downloaded on the 26.10.2016 (change date of the file) and probably installed on the same day.

Abestanis:
> I just installed the latest redistributables so I don't know whether the information is still of use to you

Unfortunately not. I tried reproducing the problem in a fresh Windows 10 Vm with only the [`Visual C++ Redistributable for Visual Studio 2015 Update 1`](https://www.microsoft.com/en-US/download/details.aspx?id=49984) installed. For some reason I couldn't find the download for  Update 2). I was unable to reproduce the problem, Birdtray started without the error message about a missing dll.

It might be either specific to Windows 7 or that Update 2. I guess I will have to create a Windows 7 Vm when I have the time to test this further.

If you want to help you could uninstall Birdtray and the latest redistributables, install the 2015 Update 2 if you still have the installer and then install Birdtray and see if you can reproduce the problem again.

jejo86:
I could reproduce the error following your instructions.

[20200427_BirdtrayIssue281.zip](https://github.com/gyunaev/birdtray/files/4541594/20200427_BirdtrayIssue281.zip)

I may have named the folder "2015 Update 2" instead of "2015 Update 1" by accident. Anyways the version seems to be 14.0.23026.0.

I could also provide the installation files of the VS C++ redistributables, if needed.

Maybe we should not put too much effort into solving this, since the workaround is easy and Win7 is deprecated anyways.

PR #285, opened 2020-03-23
Use monospace font for log window

dvalter:
It may improve log readability

### DIFF
--- a/src/dialoglogoutput.cpp
+++ b/src/dialoglogoutput.cpp
@@ -8,6 +8,7 @@ DialogLogOutput::DialogLogOutput(QWidget *parent) :
 
     // Auto-delete this object when closed
     setAttribute( Qt::WA_DeleteOnClose );
+    logViewer->setFont(QFontDatabase::systemFont(QFontDatabase::FixedFont));
 }
 
 DialogLogOutput::~DialogLogOutput()


Abestanis:
I like the idea, but doing it like this also changes the font of the `Close` button. If you add the following line to the constructor of the `DialogLogOutput` class instead, we can avoid that:
```c++
logViewer->setFont(QFontDatabase::systemFont(QFontDatabase::FixedFont));
```

dvalter:
Thank you, that's 100% a better solution. I've overlooked both "Close" button font and the reason why Monospace looked better than `QFontDatabase::FixedFont` (it was on my side).

gyunaev:
You can just change the Font property of text editor window in the UI editor. No need to write code at all.

Abestanis:
> You can just change the Font property of text editor window in the UI editor.

I tried that and couldn't get it to work. I was only able to set the font family to one of the fonts I had installed on my system.

gyunaev:
If you set it to a monospace font, and it didn't find the exact one, it should fall back into another monospace font.

Abestanis:
That is not what happened for me. It fell back to the default font, which is obviously not a monospace font.

gyunaev:
Strange, it does so on Linux.

PR #286, opened 2020-03-23
Add options for console log output

dvalter:
Add -l/--log CLI option (higher priority) and checkbox to choose
where to write Birdtray logs.

In some cases (i.e. debugging or automated log collection) having log written to stdout, stderr or file may be significantly more convenient.

### DIFF
--- a/src/birdtrayapp.cpp
+++ b/src/birdtrayapp.cpp
@@ -56,11 +56,11 @@ BirdtrayApp::BirdtrayApp(int &argc, char** argv) : QApplication(argc, argv)
         return;
     }
     
-    // This also initializes log
+    Log::initialize(commandLineParser.value("log"));
+
     Log::debug( "Birdtray version %d.%d.%d started", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH );
 
     ensureSystemTrayAvailable();
-    
     // Load settings
     settings = new Settings();
     if (commandLineParser.isSet("reset-settings")) {
@@ -68,6 +68,7 @@ BirdtrayApp::BirdtrayApp(int &argc, char** argv) : QApplication(argc, argv)
     } else {
         settings->load();
     }
+
     if (!translationLoadedSuccessfully) {
         Log::debug("Failed to load translation for %s", qPrintable(QLocale::system().name()));
     }
@@ -167,6 +168,7 @@ void BirdtrayApp::parseCmdArguments() {
             {{"s", SHOW_THUNDERBIRD_COMMAND}, tr("Show the Thunderbird window.")},
             {{"H", HIDE_THUNDERBIRD_COMMAND}, tr("Hide the Thunderbird window.")},
             {{"r", "reset-settings"}, tr("Reset the settings to the defaults.")},
+            {{"l", "log"}, tr("Write log to a file."), tr("FILE")}
     });
     commandLineParser.process(*this);
 }
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -99,4 +99,7 @@ void Log::add(const QString &text)
     // Add it to the open log window, if any
     if ( !mDialog.isNull() )
         mDialog.data()->add( logline );
+
+    // Log message to stderr
+    qDebug( "%s", qPrintable(logline) );
 }
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -4,20 +4,32 @@
 #include <QDateTime>
 #include <QStandardPaths>
 #include <QMessageBox>
+#include <QTextStream>
 
 #include "log.h"
 #include "dialoglogoutput.h"
 
 Log * Log::mLog;
 
-Log::Log()
+void Log::initialize(const QString& path )
 {
-}
+    if (mLog != nullptr) {
+        Log::fatal("Attempt to initialize initialized log");
+    }
 
-Log::~Log()
-{
-}
+    Log * self = g();
 
+    if (path == "") {
+        return;
+    }
+
+    QMutexLocker l( &(self->mMutex) );
+    self->mOutputFile.setFileName(path);
+    if (!self->mOutputFile.open(QIODevice::OpenModeFlag::Append)) {
+        l.unlock();
+        Log::fatal(QObject::tr("Failed to open log file %s").arg(path));
+    }
+}
 
 void Log::fatal( const QString& str )
 {
@@ -76,7 +88,7 @@ void Log::showLogger()
 
 Log *Log::g()
 {
-    if ( mLog == 0 )
+    if ( mLog == nullptr )
         mLog = new Log();
 
     return mLog;
@@ -102,4 +114,10 @@ void Log::add(const QString &text)
 
     // Log message to stderr
     qDebug( "%s", qPrintable(logline) );
+    
+    // if log file is open, append to log file
+    if (mOutputFile.isOpen()) {
+        mOutputFile.write( (logline + "\r\n").toUtf8() );
+        mOutputFile.flush();
+    }
 }
--- a/src/log.h
+++ b/src/log.h
@@ -4,11 +4,12 @@
 #include <QMutex>
 #include <QPointer>
 #include <QStringList>
+#include <QFile>
 
 class DialogLogOutput;
 
 // Logger
-class Log
+class Log final
 {
     public:
         // Adds a log entry and terminates an app, writing the log buffer to log.txt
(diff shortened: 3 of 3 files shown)


src/log.cpp
```diff
@@ -10,22 +10,44 @@
...
+        self->mOutputFile = stdout;
+        break;
+    case LogOutput::Window:
+        self->mOutputFile = nullptr;
+        break;
+    case LogOutput::File:
+#ifdef    Q_OS_WIN
+        ::fopen_s(&self->mOutputFile, path.toLocal8Bit().constData(), "wt");
+#else  // Q_OS_WIN
+        self->mOutputFile = ::fopen(path.toUtf8().constData(), "w");
```
gyunaev: This has to be:

> self->mOutputFile = ::fopen(path.toLocal8bit().constData(), "w");

Otherwise it will not work on Linux when local encoding is not UTF8.

gyunaev: Also wondering whether "a" would work better here?

src/log.cpp
```diff
@@ -10,22 +10,44 @@
...
+        break;
+    case LogOutput::Stdout:
+        self->mOutputFile = stdout;
+        break;
+    case LogOutput::Window:
+        self->mOutputFile = nullptr;
+        break;
+    case LogOutput::File:
+#ifdef    Q_OS_WIN
+        ::fopen_s(&self->mOutputFile, path.toLocal8Bit().constData(), "wt");
```
gyunaev: 1. Need to check the fopen_s return value. It could set mOutputFile to nonzero and still fail; the function doc doesn't say it guarantees fp not to change unless the file actually opens.
2. https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-s-wfopen-s?view=vs-2019 has unicode support, which is more reliable than local 8 bit; I suggest using Unicode (8/16) and passing the appropriate encoding.

Unicode support is relevant here, as many usernames on non-English Windows systems will have non-English characters and paths.

gyunaev: Another - and IMHO better - option would be to use QFile, which handles all that internally. You can use it for streams too. See https://stackoverflow.com/questions/4453497/how-to-get-qiodevice-instance-for-stdin-stdout-stderr-text-streams-in-qtjambi#4453827 - this should be portable.

src/log.cpp
```diff
@@ -53,8 +75,14 @@ void Log::debug(const char *fmt, ... )
     vsnprintf( buf, sizeof(buf) - 1, fmt, vl );
     va_end( vl );
 
-    // Add the log line
-    g()->add( buf );
+    Log * self = g();
+
+    if (self->mOutputFile == nullptr) {
```
gyunaev: If we are writing into log, this has to be done in add() to avoid code duplication AND to ensure the timestamps are written to the log.

src/log.h
```diff
@@ -33,14 +43,15 @@ class Log
         void    add( const QString& text );
 
         // Internal stuff
-        Log();
-        ~Log();
+        Log()  = default;
+        ~Log() = default;
```
gyunaev: If you do this, you can also add **final** to the class def.

src/log.h
```diff
@@ -33,14 +43,15 @@ class Log
...
         Log( const Log& l) = delete;
         Log operator= ( const Log& l) = delete;
 
         // All historic log entries are stored here, guarded by mMutex
         QStringList     mEntries;
         QMutex          mMutex;
+        FILE*           mOutputFile = stderr;
```
gyunaev: As stated above, this better be QFile

src/settings.h
```diff
@@ -123,6 +123,11 @@ class Settings
          */
         bool    onlyShowIconOnUnreadMessages;
 
+        /**
+         * Whether to log to console (stderr) or use line buffer
+         */
+        bool    mLogToConsole;
+
```
gyunaev: Is there any reason we need this as a setting? I only see it useful for debugging (either by devs or users), why a setting?

src/dialogsettings.ui
```diff
@@ -935,6 +935,16 @@
               </property>
              </spacer>
             </item>
+            <item>
```
gyunaev: Allowing to change this option via settings UI is not needed. People who can access log files are able to start the app with the appropriate command-line.

src/birdtrayapp.cpp
```diff
@@ -68,6 +68,26 @@ BirdtrayApp::BirdtrayApp(int &argc, char** argv) : QApplication(argc, argv)
...
+    QString logOutput = commandLineParser.value("log");
+    if (logOutput != "") {
+        if (logOutput == "stderr") {
+            Log::setLogOutput(Log::LogOutput::Stderr);
+        } else if (logOutput == "stdout") {
+            Log::setLogOutput(Log::LogOutput::Stdout);
+        } else if (logOutput == "window") {
+            Log::setLogOutput(Log::LogOutput::Window);
+        } else {
+            Log::setLogOutput(Log::LogOutput::File, logOutput);
```
gyunaev: Better to check that the log filename could be opened right here and print the error message in console if we have the console. This is expected behavior on Linux.

gyunaev:
Thank you for your pull request! I left some notes in the above, could you please check when you have time?

Abestanis:
I wonder if wouldn't make sense to always write to `stdout` and `stderr`, as well as the log viewer. The user could then use the `--log` to write it to a log file in addition to all of the above.

I haven't run the code, but I think right Birdtray is configured to log to a file, `stdout` or `stderr`, the user can still open the log viewer, but is presented with an empty page.

gyunaev:
I also see absolutely no reason to write to stdout either, honestly. Just stderr is enough.

src/log.cpp
```diff
@@ -53,8 +75,14 @@ void Log::debug(const char *fmt, ... )
...
+        // Add the log line
+        self->add( buf );
+    } else {
```
gyunaev: We have to add to buffer even if logging into file is enabled, so the user can see the log output using a button.

src/log.cpp
dvalter: I've refactored it with QFile. However I'm not sure if it would be better to use here `unique_ptr<QFile>` rather than raw pointer or it does not align well with Qt/project code style.

src/log.cpp
dvalter: I moved it to add() as an optional branch after writing to the buffer.

src/log.cpp
gyunaev: No, if you're using QFile, no need to use a pointer. With FILE you have to use a pointer though, as its an opaque type.

src/log.cpp
```diff
@@ -99,4 +143,10 @@ void Log::add(const QString &text)
     // Add it to the open log window, if any
     if ( !mDialog.isNull() )
         mDialog.data()->add( logline );
+
```
Abestanis:
```suggestion
        QTextStream(stderr) << logline << endl;
```
I think adding this here would make sense. We always want to log to `stderr`, just as we always want to log to the log viewer.

gyunaev:
Thank you for updating the pull request. A few more changes please:

1. Make QFile a class object, not a pointer. You can use setFileName and open() to open a log file.
2. Let's have log always overwritting or always appending (up to you), no need for an option.
3. No need for storing those log options in settings, or configure them through dialog. For those users who are unfamiliar they will have log window to copy-paste, and for crashes they would receive specific instructions anyway.

Thank you again for your work!

dvalter:
>     Make QFile a class object, not a pointer. You can use setFileName and open() to open a log file.
 
I'd like to do so, but I see an edge case this code won't work that great with QFile: If setLogOutput is called for the second time to change any opened file to a file that can't be written, QFile version would close current log file first and get error later. Any possibility to move new file in place of the current one would help, but QFile does not support move assignment or some equivalent to it. However it could be achieved with raw/smart pointers to QFile.

Abestanis:
> QFile version would close current log file first and get error later.

I don't understand the problem. Can't you just open the new file first, check if we can write there and then replace the old file only if it succeeded?

dvalter:
> You can close it and open it again, or reassign it. But I don't see how setLogOutput could be called again. If you want to be sure, you can create rename setLogOutput to Log::initialize and require it to be called only once.

Since we've stripped down all the complexity behind setLogOutput, it seems fair. Thank you for an idea of how to add a reason to this one shot function.

src/log.cpp
dvalter: I'm not sure about how this aligns with an initial purpose of recently removed `--debug flag, but I'd rather have a somewhat meaningful log in stderr than have none, so I like this idea.

src/log.h
dvalter: Added

PR #295, opened 2020-03-29
Display the changelog as markdown on Qt 5.14.0+

Abestanis:
Qt version from 5.14.0 onwards support displaying markdown text.
Use this feature when compiling with an adequate Qt version and display the changelog as markdown.
Changelog with markdown with Qt 5.14.1
On older Qt versions, we convert the markdown to a pure text form as best as possible:
Changelog without markdown on older Qt versions
To take advantage of the markdown feature, we bump the newest Qt on Github Actions to 5.14.1.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         os: [windows-latest, ubuntu-latest]
         arch: [x86, x64]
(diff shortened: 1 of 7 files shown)


Abestanis:
Well, this didn't work entirely as planned:
Changelog
I guess I didn't test it enough with the parentheses. But I guess it could have been worse.
How does it look for you on Linux, @gyunaev?

Issue #297, opened 2020-03-31
Make 1.8.0 release


gyunaev:
Tag pushed, @Abestanis please test the Windows version, and we should be ok to release. Linux is fine.

Abestanis:
Shouldn't this be the 1.8.0 release? Also, can we include #296? We also promised the translators to notify them two weeks in advance. Do you want to do that?

gyunaev:
Yes, you're right, 1.8.0 it is. Updated the title.

Translators, please update the translations if needed. After April 2 until release we will no longer merge pull requests for 1.8.0 which adds new strings or changes existing ones (removal is ok though). I understand everyone is busy, so we gonna give everyone a week to come back and tell us if they need more time; we will make a decision on April 9th whether to proceed with the release.

#296 has conflicts, so yes, we should include that indeed once resolved.

Abestanis:
> #296 has conflicts, so yes, we should include that indeed once resolved.

They have been resolved.

Abestanis:
We should remove the `v1.7.0` tag and the draft release. Did you send the translators an email? Otherwise we should at least `@` mention them, so they get notified.

gyunaev:
I don't have their emails. Shall we ask, or just notify here?

Abestanis:
I think it is best if we mention them here, they should get an email from Github.

gyunaev:
Pinging @Vistaus and @albanobattistella for the feedback.

Abestanis:
Since we are creating a new release soon, we might want to clean up some old stale issues. We have a few open issues that might have been fixed by changes we made since the last release, but the original reporters haven't given any feedback in months (namely #109, #129, #158, #166 and #240). Can we close those?

gyunaev:
Yep. Asked for feedback/closed.

Abestanis:
As a reminder to ourselves; We must not forget to update the version in `src/res/com.ulduzsoft.Birdtray.appdata.xml` and `contrib/rpm/birdtray.spec`.

Abestanis:
@albanobattistella Can we please get a status update from you? Do you need more time to update the Italian translation or are you not going to be able to update the translation?

Abestanis:
@gyunaev We are just missing the `1.8.0` git tag.

Abestanis:
I created the tag and tested the Windows installers. Do you want me to create the changelog as well?

gyunaev:
You can even make a release ;)

[...]

PR #301, opened 2020-04-02
Add migration from sqlite folders to mork paths

Abestanis:
This attempts to automatically migrate watched accounts settings from the `Sqlite` parser, which was removed in #181, to the mork parser. The user will be notified that their configuration was migrated and if there were any problems.

If you want to merge this, @gyunaev, we should do this today, because this adds new translation strings, which will break [our promise to not add any new translation starting from April 2 to the release](https://github.com/gyunaev/birdtray/issues/297#issuecomment-606891006).


Abestanis:
I don't know why the `Create Release` workflow is triggering and marked as failed, the branch name has no dots in it. The build did succeed though.

PR #302, opened 2020-04-03
Re add the condition to only run the 'Create Release' workflow only if a tag was added

Abestanis:
Partially revert commit 10eff850.


Abestanis:
Its not like it's even running them, they are just marked as failed. Maybe because they were skipped? Doesn't make any sense. It doesn't do it [on my fork](https://github.com/Abestanis/birdtray/actions/runs/70018876). Might be a Github bug?

PR #305, opened 2020-04-05
Add low size svg icon support

### LINKED ISSUE #303: HiDPI/svg support
On my laptop running Debian KDE the birdtray icon its always very pixelated when I try to apply a new svg icon (see picture)
I have global scaling set to 1.5
image
It it possible to fix?

dvalter:
Use QtSvg renderer to rasterize svg with an appropriate resolution
instead of upscaling a raster image later.

Fixes #303


gyunaev:
Still not building though

Abestanis:
It is building, take a look at the checks. It's the `Create Release` build that is marked as failed, even though it is skipped.

dvalter:
I don't know much about Windows, but in KDE with a relatively small tray (24x24), it still makes sense for 16x16 svg:
Screenshot_2020-04-07_01-40-52_birdtray_icon
[original icon](https://github.com/gyunaev/birdtray/files/4441229/mail-queue.svg.gz)

> Also, do you think it would make sense to let the user choose the resolution in which to import the icons?

I'm not sure about that. In most cases of svg having lower size than an actual icon on screen old variant will look ugly, otherwise there should be only a slight difference. In case of internal representation changing stored icon size from something predefined makes font rendering quite unhappy, so I'd rather avoid it. However if anyone will point out an icon that looks significantly worse when rendered to 128x128 and then downscaled, it will be a reason to add some sort of configurability here.



> That fixes it +1

Wow, interesting. On Linux I had only a gray line on the very top of some icons.

gyunaev:
From my experience, downscaled SVG icons almost always look much worse than icons created for this resolution. But up to you.

Abestanis:
> In most cases of svg having lower size than an actual icon on screen old variant will look ugly

That is actually why I am proposing a configuration here. Not to downscale but to allow using icons larger than 128x128. Will 128x128 be enough for 4k-screens and what comes beyond? I will add configuration for this, but that will have to wait until after the upcoming release.

PR #307, opened 2020-04-06
Revert "Remove Util::decodeIMAPutf7 and the "decode" command line option"

gyunaev:
Reverts gyunaev/birdtray#299


Abestanis:
Whoops, I didn't realize that the migration code uses the function.

PR #327, opened 2020-04-21

igorruckert:


src/translations/installer_de.ts
```diff
@@ -236,5 +236,9 @@
         <source>Lang_it</source>
         <translation>Italienisch</translation>
     </message>
+    <message>
+        <source>Lang_pt</source>
+        <translation>Português</translation>
```
Abestanis, line 241:
```suggestion
        <translation>Portugiesisch</translation>
```
This is the German translation file, so this should be translated in German.

PR #328, opened 2020-04-21

igorruckert:


src/translations/installer_en.ts
```diff
@@ -235,6 +235,9 @@
     <message>
         <source>Lang_it</source>
         <translation>Italian</translation>
+    </message>
+            <source>Lang_pt</source>
+        <translation>Português</translation>
```
Abestanis, line 240:
```suggestion
        <translation>Portuguese</translation>
```
This is the English translation file, so this should be translated in English.

PR #330, opened 2020-04-21

igorruckert:


src/translations/installer_nl.ts
```diff
@@ -236,5 +236,9 @@
         <source>Lang_it</source>
         <translation>Italiaans</translation>
     </message>
+    <message>
+        <source>Lang_pt</source>
+        <translation>Português</translation>
```
Abestanis, line 241:
```suggestion
        <translation>Portugees</translation>
```
This is the Dutch translation file, so this should be translated in Dutch.

Issue #304, opened 2020-04-05
Hook point for external script/program


rnc:
**Description**

This is related to #122 and #226.

I had an idea that it might be a useful 'first step' to provide a hook point to execute an external program once new mail is detected. It could pass some useful variables (like e.g. http://tjeb.nl/Projects/Mailbox_Alert/Manual/index.html does). This way users on either Windows or Linux could run their own choice of script/program once mail is detected. It would potentially be a 'quicker' method than implementing a full notification system. For instance on Linux using `libnotify` something like 
```
/usr/bin/notify-send --icon=/usr/share/icons/hicolor/48x48/apps/thunderbird.png  -a "%server : %originalfolder"  "%subject from %sendername"
```

Maboroshy:
I want that too. I'd use it to update Thunderbird icon's "badge" at the dock to show the number of unread messages. All the extensions that did it are outdated now.

gyunaev:
Committed.

Maboroshy:
Thanks. I can now use this feature to update a badge at Latte Dock by putting in

`dbus-send --dest=com.canonical.Unity --type=method_call /Latte org.kde.LatteDock.updateDockItemBadge string:"thunderbird" string:"%NEW%"` command 

or `update-badge thunderbird %NEW%` with [this shell script wrapper](https://github.com/Maboroshy/latte-dock-badge-updater).

pztrn:
For docks and taskbars that supports libunity API.

Save somewhere and `chmod +x` it:

```
#!/usr/bin/env bash

COUNT=$1

if [ ${COUNT} -gt 0 ]; then
        gdbus emit --session --object-path / --signal com.canonical.Unity.LauncherEntry.Update "application://betterbird.desktop" "{'count': <int64 ${COUNT}>, 'count-visible': <true>}"
else
        gdbus emit --session --object-path / --signal com.canonical.Unity.LauncherEntry.Update "application://betterbird.desktop" "{'count': <int64 0>, 'count-visible': <false>}"
```

Add to settings like:

```
/path/to/script.sh %NEW%
```

Issue #226, opened 2020-01-08
Use Windows notifications


kaiser-chris:
It would be nice to enable win api notifications in the windows build. This will for example allow users to use the notification center on windows 10 for missed notifications.

I understand the custom notifications are needed for platform independece and some users prefer them, but I think it would be nice to have a config option to use win api notifications.

References:

- https://docs.microsoft.com/en-us/windows/win32/shell/notification-area (Old Windows compatible)
- https://github.com/mohabouje/WinToast (lib for Windows 8/10 notifications)

Abestanis:
Birdtray doesn't send any notifications, as far as I know. If you are referring to the notifications about new Emails, those are sent by Thunderbird and unfortunately we have no control over them.

There is a [feature request for Thunderbird](https://bugzilla.mozilla.org/show_bug.cgi?id=1192615) that requests native notifications.

gyunaev:
Technically we can extract the new email info from Mork, it's there.

Abestanis:
Yes, if we want to send our own notifications. Should be possible.

rnc:
I think this is similar (from the Linux POV) to #122 

As the thunderbird native notifications appeared to be completed unreliable for me (Fedora 31) I switched to https://addons.thunderbird.net/en-US/thunderbird/addon/mailbox-alert / https://github.com/tjeb/Mailbox-Alert which allows me to configure notify-send (from libnotify) to do much better notifications.

There appears to be similar cmd line apps available for Windows to use the same pattern ( e.g. https://github.com/go-toast/toast , https://github.com/KDE/snoretoast )

PR #346, opened 2020-04-26
Hook unread counter

gyunaev:
Implemented a hook unread counter, allowing to invoke a process eachtime the unread count changes. This is for post-1.8.1 release since it adds new strings.

### DIFF
--- a/src/dialogsettings.ui
+++ b/src/dialogsettings.ui
@@ -21,7 +21,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
      </property>
      <property name="tabBarAutoHide">
       <bool>false</bool>
@@ -372,6 +372,27 @@
           </item>
          </layout>
         </widget>
+       </item>
+       <item>
+        <layout class="QHBoxLayout" name="horizontalLayout">
+         <item>
+          <widget class="QLabel" name="label">
+           <property name="text">
+            <string>Unread change cmd:</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QLineEdit" name="leProcessRunOnCountChange">
+           <property name="toolTip">
+            <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If not empty, this command will be invoked every time the unread counter changes (including when it becomes zero). It is invoked via shell, as-is, with %NEW% replaced by the new unread count value, and %OLD% replaced with the old unread count value (which may be the same as new).&lt;/p&gt;&lt;p&gt;Most users don't need this functionality and should leave it empty.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+           </property>
+           <property name="whatsThis">
+            <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If not empty, this command will be invoked every time the unread counter changes (including when it becomes zero). It is invoked via shell, as-is, with %NEW% replaced by the new unread count value, and %OLD% replaced with the old unread count value (which may be the same as new).&lt;/p&gt;&lt;p&gt;Most users don't need this functionality and should leave it empty.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+           </property>
+          </widget>
+         </item>
+        </layout>
        </item>
       </layout>
      </widget>
--- a/src/trayicon.cpp
+++ b/src/trayicon.cpp
@@ -129,6 +129,23 @@ void TrayIcon::unreadCounterUpdate( unsigned int total, QColor color )
         }
     }
     
+    // Execute the hook process
+    if ( !BirdtrayApp::get()->getSettings()->mProcessRunOnCountChange.isEmpty() )
+    {
+        QString cmdline = BirdtrayApp::get()->getSettings()->mProcessRunOnCountChange;
+
+        // Replace the %NEW% with the new unread count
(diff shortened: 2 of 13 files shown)


Abestanis:
Looks good at first glance, but I won't be able to test it until tomorrow.

> This is for post-1.8.1

What is the holdup for the release? Do you want to add something before releasing?

gyunaev:
That´s fine, we are not in hurry. No, just waiting for your translations :)

Abestanis:
> No, just waiting for your translations

Which translations? Do you mean the Spanish translations?

gyunaev:
There are a bunch of pull reqs open with translations. Will we close them before 1.8.1?

Abestanis:
No, I don't think it is worth waiting for them.
You can have a look at the update to the Russian translation if you want.

src/dialogsettings.ui
```diff
@@ -373,6 +373,27 @@
...
+          <widget class="QLabel" name="label">
+           <property name="text">
+            <string>Unread change cmd:</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QLineEdit" name="leProcessRunOnCountChange">
+           <property name="toolTip">
+            <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If not empty, this command will be invoked every time the unread counter changes (including when it becomes zero). It is invoked via shell, as-is, with %NEW% replaced by the new unread count value, and %OLD% replaced with the old unread count value (which may be the same as new).&lt;/p&gt;&lt;p&gt;Most users don't need this functionality and should leave it empty.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
Abestanis, line 388: I would remove the `Most users don't need this functionality and should leave it empty.`
Maybe it is a sign that this belongs in the `Advanced` tab? Or alternatively, leave it here but make it invisible unless enabled via a `Show advanced options` option in the `Advanced` tab?

src/dialogsettings.ui
```diff
@@ -373,6 +373,27 @@
          </layout>
         </widget>
        </item>
+       <item>
+        <layout class="QHBoxLayout" name="horizontalLayout">
+         <item>
+          <widget class="QLabel" name="label">
+           <property name="text">
+            <string>Unread change cmd:</string>
```
Abestanis, line 381:
```suggestion
            <string>Unread change command:</string>
```
We always used command in user visible stings. I am obviously not a native speaker, but `command` sounds nicer than `cmd` to me.

src/trayicon.cpp
```diff
@@ -129,6 +129,23 @@ void TrayIcon::unreadCounterUpdate( unsigned int total, QColor color )
...
+    // Execute the hook process
+    if ( !BirdtrayApp::get()->getSettings()->mProcessRunOnCountChange.isEmpty() )
+    {
+        QString cmdline = BirdtrayApp::get()->getSettings()->mProcessRunOnCountChange;
+
+        // Replace the %NEW% with the new unread count
+        cmdline.replace( "%NEW%", QString::number( total ) );
+
+        // Replace the %OLD% with the old unread count
+        cmdline.replace( "%OLD%", QString::number( mUnreadCounter ) );
```
Abestanis, line 141: What do you think about adding %DIFF% as well? Should be really easy to add and might make it easier for some commands.

Maboroshy:
> One observation though: When Birdtray is started and we detect unread emails, we also run the hook. This feels like it is undesired behavior, I guess many people want to use that hook to detect new mails and not the old ones that have been sitting in their Inbox for 5 months (definitely not talking about my Inbox here wink)

I want to use this feature to update the unread counter on the icon's badge at dock. So running the hook at start looks like a desired behavior for my use case. That way I can go without any counter storage between sessions. Maybe there can be an extra checkbox to make it optional.

Maboroshy:
I've tested this branch on Manjaro Linux. All good. Managed to update Latte Dock badge counter with a tiny Python script I'm going to publish after this feature is released.

rnc:
Is this for #304 ? 
Would it be possible to also pass e.g. account causing the count change ? 

Does the parser have knowledge of the folder (and even the sender/subject) as well which could be passed through?

gyunaev:
Of folder, yes. But not of the account. Note that many files could have changed.

rnc:
> Of folder, yes. But not of the account. Note that many files could have changed.

@gyunaev Thanks for the reply. Could I clarify that please - when I choose `Accounts to Monitor` it displays `Profile/Server(Account)/Folder` e.g. `Profile2020 / imap.gmail.com / Bugzilla `. Therefore surely if it knows about the Bugzilla folder, it should know about the imap.gmail.com account? 

Yes I understand its feasible given the timings that multiple emails may have arrived. What about something like sending to the external program 
```
<numericCountChange> [ <server>:<folder ].....n
```
 i.e. send a list of server , separator, folder as  a list of parameters ?

gyunaev:
Too much work. You're welcome, of course, to implement it and submit a pull request.

rnc:
@gyunaev Am I following it correctly with my supposition regarding server/folder knowledge?

Are you able to pass any further information in the interim? I haven't done C++ in decades so would have to see if I have any time to dig out my old books on it ;-)

Abestanis:
> Am I following it correctly with my supposition regarding server/folder knowledge?

Yes, we do have that information. We will definitely be able to pass more information to the hook, but I agree with @gyunaev, let's first get the basics going. We will also need all of that information for #226 and #163. I did start working on that a few moths ago but wasn't able to finish it. Once I do we can add all of the information to the hook.

PR #348, opened 2020-04-27
Require qt5-qttranslations

raven-kg:
https://github.com/gyunaev/birdtray/issues/333
Added qt5-qttranslations as a weak dependency

PR #350, opened 2020-04-28
Resolve building RPM within Fedora COPR

rnc:
Some minor fixes to resolve building directly from the github release archive in order to build https://copr.fedorainfracloud.org/coprs/rnc/Birdtray/

### DIFF
--- a/contrib/rpm/birdtray.spec
+++ b/contrib/rpm/birdtray.spec
@@ -4,11 +4,11 @@ Release:      1%{?dist}
 Epoch:        0
 License:      GPLv3
 Group:        System Environment/Shells
-Source0:      https://github.com/gyunaev/%{name}/archive/RELEASE_%{version}.tar.gz
+Source0:      https://github.com/gyunaev/%{name}/archive/%{version}.tar.gz
 URL:          https://github.com/gyunaev/birdtray
 BuildRoot:    %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildRequires: cmake3 gcc-c++ desktop-file-utils
-BuildRequires: sqlite-devel qt5-qtbase-devel qt5-qtx11extras-devel qt5-qtsvg-devel
+BuildRequires: sqlite-devel qt5-qtbase-devel qt5-qtx11extras-devel qt5-qtsvg-devel qt5-linguist
 
 #%if 0%{?fedora} || 0%{?rhel} >= 8
 #Recommends:      thunderbird
@@ -20,7 +20,7 @@ Summary: Birdtray is a free system tray notification for new mail for Thunderbir
 Birdtray is a system tray new mail notification for Thunderbird, which does not require extensions.
 
 %prep
-%setup -q -n %{name}-RELEASE_%{version}
+%setup -q -n %{name}-%{version}
 
 %build
 mkdir %{_target_platform}


contrib/rpm/birdtray.spec
```diff
@@ -4,11 +4,11 @@ Release:      1%{?dist}
 Epoch:        0
 License:      GPLv3
 Group:        System Environment/Shells
-Source0:      https://github.com/gyunaev/%{name}/archive/RELEASE_%{version}.tar.gz
+Source0:      https://github.com/gyunaev/%{name}/archive/%{version}.tar.gz
 URL:          https://github.com/gyunaev/birdtray
 BuildRoot:    %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildRequires: cmake3 gcc-c++ desktop-file-utils
-BuildRequires: sqlite-devel qt5-qtbase-devel qt5-qtx11extras-devel qt5-qtsvg-devel
+BuildRequires: sqlite-devel qt5-qtbase-devel qt5-qtx11extras-devel qt5-qtsvg-devel qt5-linguist
```
Abestanis, line 11: #348 proposes to add this as an optional dependency.
@raven-kg, what do you think?

rnc: Thats fine, if #348 is merged first then I can rebase. If `qt5-linguist` isn't installed I think the cmake needs more if checks for translations

rnc: Actually ignore that last bit - qt5-linguist should still be installed as a result of 348.

Abestanis: It is probably best if you just include #348 here. I would like to include this in the `1.8.1` release and #348 has a merge conflict anyway, so lets just speed up the process.

raven-kg: @rnc thanks for the fixes. During the build process for different versions of CentOS I had to make several fixes that made #348 irrelevant, so I should roll back that commit and push new one. So, I'll also pull your changes) There is only one thing that I should ask

Just one thing I'd like to ask - is it better to write a url for 'Source0' in such a way to get a named archive?
`Source0:      https://github.com/gyunaev/%{name}/archive/%{version}.tar.gz#/%{name}-%{version}.tar.gz`

rnc: @raven-kg I'm a little bit confused on your plan. Do you mean to for @Abestanis to merge this, then you layer upon further fixes?

Regarding your question about the naming, if you use an archive fragment won't that mean it is expecting an archive named `birdtray-1.8.0.tar.gz`? 

Having looked back through the releases I now understand why you used `RELEASE_` ; I think the naming needs to be tied into to whatever name format is used in the github asset deployment - in which 1.8.0 changed the format - @Abestanis are there any plans to change the format again?

raven-kg: Yes, I think @Abestanis should merge your changes first. 
I use mock and my own hands to build rpms, so every time I download an archive using "wget -O birdtray-1.8.0.tar.gz" I expect to get a readable archive name instead of 1.8.0.tar.gz, I guess there's nothing wrong with that)

rnc: The COPR I setup expected the Sources URL to lead to a valid url - it needs to correspond to the release process so the assets on github match.

raven-kg: @rnc ok, let's keep your version.

Abestanis: > @Abestanis are there any plans to change the format again?

Yes, we want to move to the `vMAJOR.MINOR.PATCH` format that is recommended by GitHub, but we had to do it in two steps, because the auto update detection in Birdtray also depends on the format. I explained this a bit in #298. When we change it again I will make sure to update the spec file, I just wasn't aware that it depended on the format before.
My goal is to automate the release process as much as possible so we will be able to enforce the format.

Abestanis: Ups, I just realized that you @raven-kg  want to add `qt5-qttranslations` as an optional dependency, not `qt5-linguist` in #348. My bad, sorry for the confusion.

Issue #46, opened 2018-12-03
Autostart not working


yoshegg:
Using Ubuntu 18.10, birdtray won't start if put into startup applications. 

The following message can be found in `/var/syslog`:
`Sorry, system tray cannot be controlled through this addon on your operating system`

Maybe birdtray is started before the gnome-shell's system tray is loaded?

yoshegg:
A workaround is to use the following command as startup command: 
`bash -c "sleep 20 &&  /usr/bin/birdtray"`

vedgy:
Hi @gyunaev,
I just got this error message on my Manjaro Xfce desktop after large system upgrades. I think that *birdtray* doesn't wait sufficiently long for the system tray to become available due to the following mistake in code:
according to the comment at the top of [`ensureSystemTrayAvailable()`](https://github.com/gyunaev/birdtray/blob/18623475d6f20e2ecbd7aa8b4e91032fe0b535a2/src/main.cpp#L15), this function is supposed to wait for up to 60 seconds. But it calls [`QThread::usleep( 500 )`](https://doc.qt.io/qt-5/qthread.html#usleep) for up to 120 times. (500 microseconds) * 120 = 60 milliseconds. I think that [`QThread::msleep( 500 )`](https://doc.qt.io/qt-5/qthread.html#msleep) should be called instead, though I haven't actually tried or tested such a change.

gyunaev:
Yes, correct. Fixed.

PR #384, opened 2020-06-25
Sys tray icon is automatically added by QSystemTrayIcon (Fixes #357)

### LINKED ISSUE #357: Birdtray does not start
**OS**: Arch 
**Birdtray version**: I can not open the app. I have installed birdtray from the AUR 
**Thunderbird version**:68.7 
**Birdtray origin**: AUR 
**Qt version**: _NA_ 

**Description**

I have installed birdtray on GNOME Wayland, tried to start it, but nothing happened.
After 1-2 minutes, I got an error message, that the status/system bar can not be be controlled by this addon. It also says that a protocol were written in the file /tmp/log.txt, but there is no with that name.

**To Reproduce**

1. Install birdtray from the aur on GNOME Wayland.
2. Try to start it

Edit:
Tried to start it on GNOME Xorg. Still the same issue.

neoh4x0r:
Birdtray fails to start while waiting for the SystemTray to become available.

Per the QT documentation for QT4 & 5: https://doc.qt.io/qt-5/qsystemtrayicon.html#isSystemTrayAvailable	 

The system tray icon will be automatically added if the system tray is visible (even if it was
not available at launch).

Fixes issue #357

### DIFF
--- a/src/birdtrayapp.cpp
+++ b/src/birdtrayapp.cpp
@@ -62,7 +62,6 @@ BirdtrayApp::BirdtrayApp(int &argc, char** argv) : QApplication(argc, argv)
 
     Log::debug( "Birdtray version %d.%d.%d started", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH );
 
-    ensureSystemTrayAvailable();
     // Load settings
     settings = new Settings();
     if (commandLineParser.isSet("reset-settings")) {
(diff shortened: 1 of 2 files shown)


neoh4x0r:
It would seems that this does not completely fix the issue.

If I login to cinnamon (after a system restart) cinnamon will take a while to show the desktop and system tray.
When this happens birdtray is running (with this PR), but no system tray icon is showing.

If I logout of cinnamon and then log back in, birdtray is launched (and the system tray icon is shown).

I will push more commits to this PR once I debug the actual problem.

gyunaev:
This seem to be Wayland-specific issue, because this code is necessary on other desktop managers when Birdtray fails to start; see for example https://github.com/gyunaev/birdtray/issues/46

Can you make this Wayland-only?

neoh4x0r:
> This seem to be Wayland-specific issue, because this code is necessary on other desktop managers when Birdtray fails to start; see for example #46
> 
> Can you make this Wayland-only?

I don't know how to make it wayland-only...

There seems to be some issue that occurs at login after a restart....but does not occur after logging in once, then doing a logout/login. 

I believe that this PR needs to be closed since it does not seem to address the actual issue.

Also the 60 second timeout might not be enough for some users (for me it might take 120 seconds before I see the desktop and taskbar).

PR #406, opened 2020-09-29
Update main_pt.ts

igorruckert:


src/translations/main_pt.ts
```diff
@@ -433,23 +433,23 @@ OpenSSL não deve estar instalado.</translation>
     </message>
     <message>
         <source>Unread change cmd:</source>
-        <translation type="unfinished"></translation>
+        <translation>Comando para mudar não lida:</translation>
     </message>
     <message>
         <source>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If not empty, this command will be invoked every time the unread counter changes (including when it becomes zero). It is invoked via shell, as-is, with %NEW% replaced by the new unread count value, and %OLD% replaced with the old unread count value (which may be the same as new).&lt;/p&gt;&lt;p&gt;Most users don&apos;t need this functionality and should leave it empty.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</source>
         <translation></translation>
```
Abestanis, line 440: It looks like this translation is also missing. Do you want to add it?

Abestanis (changes requested):
Hi, thank you for updating the translations.
The new ones look good, but there is one missing and I found two problems with the existing translations:
* In the `installer_pt.ts` at line `189`, there is a `$` missing in front of the `{PRODUCT_NAME}`.
* Similar, in the `installer_pt.ts` at line `213`, there is a `$` missing in front of the `{ARCH}`.

Could you please fix these? Thank you.

src/translations/main_pt.ts
igorruckert: Fixes done.

Abestanis (approved):
Awesome, thank you for the quick response and for the updates to the translations.

PR #409, opened 2020-09-29
Add linguistic problem matchers

Abestanis:
This allows GitHub Actions to produce comments like the following on pull request:

Error comment by GitHub actions on pull request

It might make it easier for translators to spot errors in their files and correct them without requiring a review from us.


gyunaev (commented):
Looks good to me, please go forward.

PR #412, opened 2020-10-04
Fix never writing the logs on a fatal error

Abestanis:
The log file was never opened. Also change the log file name to be a bit more unique and easy to find.

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -49,14 +49,14 @@ void Log::fatal( const QString& str )
     self->add( str );
 
     // Write the whole log buffer into a file log.txt
-    QFile file( QStandardPaths::writableLocation( QStandardPaths::TempLocation ) + QDir::separator() + "log.txt" );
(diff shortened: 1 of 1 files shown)


src/log.cpp
```diff
@@ -49,14 +49,14 @@ void Log::fatal( const QString& str )
     self->add( str );
 
     // Write the whole log buffer into a file log.txt
-    QFile file( QStandardPaths::writableLocation( QStandardPaths::TempLocation ) + QDir::separator() + "log.txt" );
+    QFile file( QStandardPaths::writableLocation( QStandardPaths::TempLocation ) + QDir::separator() + "birdtray-log.txt" );
```
gyunaev, line 52: So we also rename the log file?

Abestanis: Yes, this is the new file we create to dump the logs in case of a fatal error. This is not the file that can be configured on the command line. For me, there were many files with `log` in my temp folder, so adding the `birdtray-` prefix helped in finding the log file.

PR #423, opened 2020-10-11
Fix codeql-analysis workflow

Abestanis:
Looks like the Autobuild action is not compatible with CMake. We fix this by copying the relative steps for Ubuntu from the Build workflow.


gyunaev:
Thank you!

PR #424, opened 2020-10-12
Enforce vX.X.X as the version pattern for the git tag

Abestanis:
As discussed in #298 I would like for us to use `vX.X.X` as the format for release git tags. This is the recommended and most commonly used format and it helps to prevent false positives in the GitHub Actions workflow that checks if it should create release files.

@rnc Could you please verify that my changes to the `spec` file are correct?
@gyunaev Please merge this before creating the new release. When you create the new release, please use `v1.9.0` as the git tag.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,7 +96,7 @@ jobs:
         id: check-deploy
         shell: bash
         run: |
-          if [[ '${{ github.event.ref }}' =~ ^refs/tags/[0-9]+\.[0-9]+\.[0-9]+$ && "$RUNNER_OS" == "Windows" ]]; then
+          if [[ '${{ github.event.ref }}' =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$ && "$RUNNER_OS" == "Windows" ]]; then
             echo ::set-output name=is_deploy::true
             echo "Deployment build detected"
           else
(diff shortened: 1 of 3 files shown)


Abestanis:
@raven-kg Could you also please verify that my changes to the spec file are correct?
@rnc Sorry, I just realized that you aren't the one who added the file in the first place. But it looks like you know how it works, so I'd still appreciate your feedback if you have the time.

raven-kg:
@Abestanis the link looks wrong. 
https://github.com/gyunaev/birdtray/archive/v1.8.1.tar.gz returns 404. As far as I can see [here](https://github.com/gyunaev/birdtray/releases/tag/1.8.1), actual link has no 'v' sign before version number

Abestanis:
> As far as I can see here, actual link has no 'v' sign before version number

@raven-kg That is true for the old releases. I'm trying to switch the format (hopefully for the last time) for all new releases in the future, starting with the upcoming `1.9.0` release.

Abestanis:
@raven-kg To clarify my question to you is: When we release the new version and upload the artifact at https://github.com/gyunaev/birdtray/archive/v1.9.0.tar.gz (and replace `1.8.1` with `1.9.0` in the spec file), will the new script correctly handle the `v`.

raven-kg:
@Abestanis sorry for the late reply. Yes, it works fine for me, but also it would be great to hear something from @rnc (he mentioned that this spec file used by koji or something similar to build packages automatically).

Abestanis:
@raven-kg Thanks for having a look.
@gyunaev I think you can merge this and create the new release now, we have been holding it up long enough now. If @rnc finds something wrong with the spec file we can fix it after the release like we did in #350.

gyunaev:
If you have time, please create a release once Windows binaries are ready, and you can give them a cursory look, thanks!

PR #428, opened 2020-10-27
Improve the translations

Abestanis:
This pull request improves the translations in a few ways:

* Move translations from the `QCoreApplication`, `QApplication` and `QObject` context into more descriptive contexts.
* Reword some of the English texts, make sure that each sentence displayed in a tool-tip ends with a dot.
* Improve the translation checker:
    - Now also runs for the Spanish translation.
    - Can handle invalid XML better now.
    - Now outputs file and line information in a more human readable format, so it can be used locally on the command line more easily.
* Ignore `unfinished_translation` warnings, to keep the focus on the real warnings.
* Add some additional checks to verify translation filenames and registration in `CMakeLists.txt`.
* Specify the correct line endings for translation files in the `.gitattributes`. The `lupdate` tool always uses `LF`, even if it is run on Windows. On the other hand, the installer translations are for Windows, so they should be `CRLF`.

@gyunaev As this changes C++ code and some of the English source strings I would appreciate if you could have a look at it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+* text=auto
+src/translations/main_*.ts text eol=lf
+src/translations/dynamic_*.ts text eol=lf
+src/translations/installer_*.ts text eol=crlf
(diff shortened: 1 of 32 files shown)


gyunaev (approved):

gyunaev:
Merge when ready, please.

Issue #313, opened 2020-04-17
export ENV var before launch thunderbird


asarubbo:
**Description**

Would be great if there will be something that let you export some env variables.
The need comes from the fact that if you are using a dark theme, thunderbird will be unreadable, so I'm exporting GTK_THEME before launch thunderbird
Thanks

Abestanis:
I don't know if that is worth doing in Birdtray. Couldn't you use a wrapper shell script to archive that?

asarubbo:
Hi, sure I can use a wrapper and this is something I'm already doing. But if this is something done by a lot of people it may worth..

gyunaev:
You don't even need a wrapper, just export those variables before starting birdtray. It should inherit the environment and pass it to children processes.

gudvinr:
This is really useful though. You can just use `env GTK_THEME=theme thunderbird` from `.desktop` file (or application menu entry) but you can't do that from birdtray so you should introduce different launch script or make copy of `.desktop` file to use custom script which sounds just like workaround but not a handy solution.

Abestanis:
> So only necessary thing to solve this issue is to correctly handle launch string because it seems like right now birdtray treats it like whole path to thunderbird executable rather than command.

@gudvinr That doesn't sound right, did you actually test that? The last time I checked, we did support entering a full command in the advanced tab.

gudvinr:
@Abestanis how would I know that otherwise?

Yes, I did test that and it didn't work. I get an error:
```
Error starting Thunderbird as '/home/{username}/env GDK_DPI_SCALE=0.5 /usr/bin/thunderbird':

execvp: No such file or directory
```

If I use launch command as `env GDK_DPI_SCALE=0.5 /usr/bin/thunderbird`.

But it works if I run `/usr/bin/env GDK_DPI_SCALE=0.5 /usr/bin/thunderbird` though it is neither convenient nor obvious.

Abestanis:
> how would I know that otherwise?

Sorry, I'm not a native speaker, do you mean how you could verify that it works or not? You seem to have figured it out.

> If I use launch command as env GDK_DPI_SCALE=0.5 /usr/bin/thunderbird

This doesn't work, because we don't search for the program on the `PATH`. You need to specify the full path to the program.

> But it works if I run /usr/bin/env GDK_DPI_SCALE=0.5 /usr/bin/thunderbird

Exactly like this. 😄👍 

> though it is neither convenient nor obvious.

[...]

PR #434, opened 2020-11-01
Add Qt 5.15.1 compatibility

Abestanis:
This pull request adds support for Qt `5.15.1`.
There is still a big problem that we need to discuss:

`QProcess::startDetached(command)` has become deprecated and might be removed in future Qt versions. We need to use `QProcess::startDetached(executable, arguments[])`. This prevents us from passing in the whole shell command as a whole, we are required to at least split the executable path from the arguments. This is especially problematic for the new unread counter change hook, because we allow entering the full shell command: https://github.com/gyunaev/birdtray/blob/9ff57bb54c8e09c0a4300d4906e1b09ccd184b8c/src/trayicon.cpp#L136-L144

We need to be able to differentiate between the executable and the arguments. I propose adding a custom widget like this:

The proposed new widget

A click on the configure button opens a dialog that allows to configure arguments and environment variables. I based the dialog on UI elements from the JetBrains Products, which I find very intuitive to use. This would also allow us to implement #313. We could use the same widget in the `Advanced` settings tab for the Thunderbird command.

@gyunaev What are your thoughts on this? I know you want to keep things as simple as possible and I think I can implement this without to much complexity. The alternative would be to just use `Utils::splitCommandLine` and hope the user doesn't input anything crazy.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,14 +10,14 @@ jobs:
       matrix:
         os: [windows-latest, ubuntu-latest]
         arch: [x86, x64]
-        qt_version: [5.9.9, 5.14.1]
+        qt_version: [5.9.9, 5.15.1]
         include:
           - os: windows-latest
             arch: x86
-            qt_compile_suite: win32_msvc2017
+            qt_compile_suite: win32_msvc2019
           - os: windows-latest
             arch: x64
-            qt_compile_suite: win64_msvc2017_64
+            qt_compile_suite: win64_msvc2019_64
         exclude:
           # We only want to test for the latest version of Qt on Windows
           - os: windows-latest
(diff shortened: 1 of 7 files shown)


gyunaev:
No, not worth changing anything simply because it is deprecated. When/if it gets removed, then we can see. They could add another replacement.

gyunaev (commented):
What benefit does it introduce?

gyunaev:
What I do to get rid of annoying deprecated warnings, -W-no-deprecated

ladar:
I know this is an old PR, but I'm wondering if any of the changes it introduces would fix #500 ...

@nicolamori can you try compiling this outdated PR and see if works?

Abestanis:
I'm going to close this for now, we can always revive it if we decide that we need a newer Qt combability.
Regarding the deprecated `startDetached`, this function is removed in Qt6 and the only alternatives are to pass the program and arguments separately. So if we ever want to have Qt6, we'll have to tackle that problem, but we'll figure it out when we get there.

PR #445, opened 2021-01-07
French translation

canoine:
Note : I didn't check this translation into birdtray.


Abestanis (approved):
Awesome, looks good! 👍
Thank you very much.

PR #448, opened 2021-01-15
Don't test compilation with Qt5.6 anymore

Abestanis:
Support for Qt5.6 ended in March 2019. The Qt team has now removed it from the [online repository](https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/), so it's no longer possible to automatically download it on the GitHub Actions runner.

This fixes the failing build.

The next course of action is to raise the minimum Qt version to 5.9. Is that ok with you, @gyunaev?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         os: [windows-latest, ubuntu-latest]
         arch: [x86, x64]
-        qt_version: [5.6, 5.9, 5.14.1]
+        qt_version: [5.9, 5.14.1]
         include:
           - os: windows-latest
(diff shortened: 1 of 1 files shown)


gyunaev:
For building we can test on 5.9+. For deployment I'd keep 5.6 as long as we're not missing out major features.

Abestanis:
Ok, but I just like to mention that without testing, it is easy to use a function that was added after Qt5.6, I did that once already. Also, the Qt devs removing that version from the repository to me is a clear sign that no one should be using it anymore.

gyunaev:
Yeah, that's possible, but we'll try to fix it then if reported. There will always be some Debian using it until 2024.

PR #452, opened 2021-01-16
Add option to ignore unread emails on showing Thunderbird

### LINKED ISSUE #447: [Feature Request] Reset unread emails (ignore) count when showing Thunderbird from tray
In current version there is setting **When clicking on Birdtray icon to hide Thinderbird, reset the icon by ignoring all currently unread emails**. Which works -> Thank you!

I'd like to have the same option when we click tray icon to **Show** Thunderbird. Context here -> https://github.com/gyunaev/birdtray/issues/377#issuecomment-754957703

Maybe additional checkbox (or radio?) with options **...reset the icon ... when clicking**:
- [ ] hide Thunderbird
- [ ] show Thunderbird

Also maybe also text in tray icon is not resetting correctly?
I can see this even after icon is resetted fine:
image

My current options are:
image

Thank you in advance for considering this.

Abestanis:
This adds the option to ignore all currently unread emails when showing the Thunderbird window.
It also moves all ignored emails related settings into one tab (removes the `When clicking on Birdtray icon to hide Thunderbird, reset the icon by ignoring all currently unread emails` option in the `Hiding` tab):

The monitoring settings tab with the new options

Further changes:
* The number of ignored emails is no longer reset to 0 if an email is marked as read and the unread email count drops below the original ignored email count. Instead, the number of ignored unread emails is simply decreased to match the number of currently unread emails.
* When Thunderbird minimized and then hidden, the `ignore unread emails when hiding` option will work now.
* Fixes a bug where the context menu would not display the number of ignored mails after leaving the settings dialog.
* The ignored email count and the `Show\Hide Thunderbird` menu entries are no longer updated, if the windowtools fail to hide/show a window.
* Fixed tab ordering in the settings dialog.

@gyunaev Can you please make sure that the changes in the X11 windowtools work on Linux?

Closes #447.

### DIFF
--- a/src/dialogsettings.cpp
+++ b/src/dialogsettings.cpp
@@ -76,10 +76,11 @@ DialogSettings::DialogSettings( QWidget *parent)
     spinUnreadOpacityLevel->setValue( settings->mUnreadOpacityLevel * 100 );
     spinThunderbirdStartDelay->setValue( settings->mLaunchThunderbirdDelay );
     boxShowUnreadCount->setChecked( settings->mShowUnreadEmailCount );
-    ignoreStartupMailCountBox->setChecked(settings->ignoreStartUnreadCount);
+    ignoreMailCountOnStartupBox->setChecked(settings->ignoreUnreadCountOnStart);
+    ignoreMailCountOnShowBox->setChecked(settings->ignoreUnreadCountOnShow);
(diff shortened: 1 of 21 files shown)


gyunaev (commented):
Looks good to me.

PR #454, opened 2021-01-18
Improve how the list of watched mork files is stored at runtime

Abestanis:
This merges the `mFolderNotificationColors` and `mFolderNotificationList` into one variable called `watchedMorkFiles`; an ordered map.
This avoids inconsistencies between the two and fixes a potential bug where we would duplicate each watched mork file in the `mFolderNotificationList` if we reloaded the settings (which is not possible at the moment, so not a problem yet).

I also added tests for the OrderedMap.
If you prefer it, I can move the OrderedMap into a new header file.

### DIFF
--- a/src/modelaccounttree.cpp
+++ b/src/modelaccounttree.cpp
@@ -13,9 +13,9 @@ ModelAccountTree::ModelAccountTree(QObject* parent, QTreeView* treeView)
         : QAbstractItemModel(parent), QStyledItemDelegate(parent) {
     Settings* settings = BirdtrayApp::get()->getSettings();
     // Get the current settings in proper(stored) order
-    for (const QString& path : settings->mFolderNotificationList) {
+    for (const QString& path : settings->watchedMorkFiles.orderedKeys()) {
         mAccounts.push_back(path);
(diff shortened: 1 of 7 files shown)


gyunaev:
The problem with this approach is that it breaks the order of folders selected by the user. When a user adds a new folder now, this folder always goes to be the last. If you use the map, the new folder can appear anywhere in the list. This is confusing a best.

If you want to use a single array, I'd still use the list, but store the QPair< QString, QColor> there.

Abestanis:
> If you use the map, the new folder can appear anywhere in the list.

That is strange, I designed the OrderedMap to prevent exactly that, to preserve the order of insertion. Can you tell me how you got it to insert randomly?

gyunaev:
> > If you use the map, the new folder can appear anywhere in the list.
> 
> That is strange, I designed the OrderedMap to prevent exactly that, to preserve the order of insertion. Can you tell me how you got it to insert randomly?

Ah you're right, I see now. Still don't see any advantage of QList< QPair< QString, QColor > > except a faster lookup. But won't hurt.

PR #546, opened 2023-04-16
Add a setting to change the searched Thunderbird process name on Windows

### LINKED ISSUE #545: Betterbird support
**Description**

I am currently evaluating a switch from Thunderbird to the fork [Betterbird](https://www.betterbird.eu) on a Windows 10 (22H2) system and the only thing I miss from running Thunderbird with the still vastly superior to native birdtray, is that birdtray doesn't seem to understand that Betterbird exists.
I tried modifying the "command line" and the "window name pattern" to match the Betterbird installation, but that changed nothing.

Is this something that is supposed to be working already and I'm just not reconfiguring some component as required, or is this a new feature that would have to be implemented?
In the latter case, I am requesting that, now, here.
I suppose the "Monitoring"-tab would have to be reconfigured to the new Profiles-directory, but since the icon remains stuck at the icon for "Thunderbird is not running" I'll defer reconfiguring all of those until the running-detection is compatible.
Also, once that is done, there should not be any incompatibilities with monitoring Betterbird's profile-files rather than Thunderbird's, right?

Using Betterbird or Thunderbird, now that I know how much better than native this functionality can be handled through birdtray, is just not an option without birdtray.

Abestanis:
This setting is only useful on Windows and allows to change the name of the process that is searched to detect if Thunderbird is running. Closes #545.

Changing the setting will take effect when saving, except for if Birdtray already found a process with the old settings.
In that case, Birdtray has to be restarted or the currently tracked process has to exit.
I think that is acceptable, since this is an advanced feature.

New setting

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -211,7 +211,7 @@ if(WIN32)
 
     # Find ssl libraries
     find_package(OpenSSL QUIET)
-    get_filename_component(SSL_PATH ${OPENSSL_SSL_LIBRARY} DIRECTORY)
+    get_filename_component(SSL_PATH "${OPENSSL_SSL_LIBRARY}" DIRECTORY)
     if(SSL_PATH)
         STRING(REGEX REPLACE "/" "\\\\\\\\" SSL_PATH ${SSL_PATH})
     else()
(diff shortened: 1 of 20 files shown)


gyunaev:
See 1ab70f2..a6da1da

Abestanis:
That is a good idea! Could you please also hide the `Thunderbird window name pattern` and `Support non-NETWM compliant WMs` on Windows, while you are there, if it is not too complicated?

Abestanis:
Ignore the remark about the `Thunderbird window name pattern` setting, I see that you have already done it. 😄

gyunaev:
Done, 1b1c7a5

PR #565, opened 2023-08-31
Fix 64 bit Windows installer not working

### LINKED ISSUE #564: On Win 10, 1.11.2 installer does not install
**OS**: Windows 10 Pro, version 22H2, build 19045.3324 (the current main), 64 bit 
**Birdtray version**: 1.11.2 (the current main, newly released) 
**Thunderbird version**: Betterbird 115.2.0-bb10, 64 bit 
**Birdtray origin**: From installer: https://github.com/gyunaev/birdtray/releases/tag/v1.11.2 
**Qt version**: _NA_ 

**Description**

Executing the installer downloaded from GitHub shows the installer language selection dialog box then just exits without displaying any further dialog windows or apparently installing any files (default location checked) **if** there was no Birdtray remaining installed when running the installer.
If another version of Birdtray (https://github.com/Abestanis/birdtray/actions/runs/4726093532) was already installed and not uninstalled, then when executing the installer the mouse cursor changes to arrow-and-hourglass ("working in background") for 1-2 seconds and then nothing more appears to happen from the execution, with the files in the default location remaining unchanged.
So, in neither case does the new Birdtray installer install Birdtray.

**Expected behavior**

The Birdtray installer is expected to install Birdtray.

**To Reproduce**

See the "Description"-section.

Abestanis:
I pinned NSI to a version which produced broken 64 bit installers. Updating the version fixes the issue.
I tested both 64 and 32 bit installers and uninstallers and found no issues.

Some of the translations got slightly changed, I took the opportunity to fix them.

I increased the version number, so after this is merged, please push a `v1.11.3` git tag, and once the CI creates the release please undraft it. Thanks.

Fixes #564.


Abestanis:
Damm, that was fast 🤣

Issue #426, opened 2020-10-25
Adding show/hide support for full Wayland


gyunaev:
At this moment Birdtray works fine with "wayland" desktop which also runs embedded X Server. However Birdtray does not work in so-called "full wayland" support, without the X Server. 

The main limitation is that Wayland provides no standard way to figure out which other top-level windows are being shown. This in turn means Birdtray cannot interact with them. This impacts the following functionality:

- Birdtray cannot detect whether Thunderbird is running, and thus cannot start/restart it;
- Birdtray cannot hide and show Thunderbird;

Those limitations will not be fixed until Wayland fully implements the [Wayland Toplevel Management protocol](https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-foreign-toplevel-management-unstable-v1.xml) which is currently unstable and not implemented by KDE or Gnome window managers. The only possible current alternative is to integrate directly with each window manager, but considering the number of them, this is not a viable alternative. Thus the full Wayland support is still pending on when this specification a) becomes stable and b) becomes implemented.

The email monitoring and showing functionality works fine, however.

gothmog123:
there aren't many working wayland window managers - some would say it's only sway and gnome. so you wouldn't have that many target if you wanted to implement it individually

sc0ttzilla:
Looks like this project is dead? BT doesn't work well at all with Thunderbird 91*. Constantly needs to be reloaded to work properly. Regardless of configuration. Any feedback on this?

gyunaev:
Please create a separate bug report.

aindriu80:
I am having the same Wayland problem, I am running Pop OS 22.04 and just switched to Wayland on the login screen.  When I logged in birdtray started up but it couldn't minimize Thunderbird, is there any way of getting birdtray to work with Wayland?

Flying--Dutchman:
For the time being, a simple workaround is to start birdtray with x11 session. Simply execute the following command to start birdtray:
`XDG_SESSION_TYPE=x11 birdtray`

fadialzammar:
I was able to get around this by launching Thunderbird as an X11 app through XWayland. Birdtray was already running under X11 for me by default, but if not you can follow @Flying--Dutchman's solution.

Modify `/usr/share/applications/thunderbird.desktop`, changing `Exec=thunderbird` (I had 3 spots in the file with this that I changed) to `Exec=env GDK_BACKEND=x11 thunderbird`.

[...]

PR #584, opened 2024-01-01
Allow showing/hiding Thunderbird when launched using Wayland platform

cptpcrd:
If the Wayland platform is in use, simply connect to the XWayland server directly, and perform queries using that connection.

As long as Thunderbird is running under XWayland, this allows Birdtray to find the window and show/hide it no matter what platform Birdtray is using.

This fixes _some_ of the problems described in e.g. #426, but it is not full Wayland support.

### DIFF
--- a/src/windowtools_x11.cpp
+++ b/src/windowtools_x11.cpp
@@ -359,6 +359,17 @@ WindowTools_X11::WindowTools_X11()
     mWinId = None;
     mHiddenStateCounter = 0;
 
+    if ( QX11Info::isPlatformX11() )
+    {
+        display = QX11Info::display();
+        root = QX11Info::appRootWindow();
+    }
+    else
+    {
+        display = XOpenDisplay(NULL);
+        root = display ? DefaultRootWindow(display) : 0;
+    }
+
     connect( &mWindowStateTimer, &QTimer::timeout, this, &WindowTools_X11::timerWindowState );
     mWindowStateTimer.setInterval( 250 );
     mWindowStateTimer.start();
@@ -370,13 +381,14 @@ WindowTools_X11::~WindowTools_X11()
 
 bool WindowTools_X11::lookup()
 {
-    if ( QX11Info::appRootWindow() == 0 )
+    // Root window is invalid (error connecting?)
+    if ( root == 0 )
         return false;
 
     if ( isValid() )
         return true;
 
-    mWinId = findWindow(QX11Info::display(), QX11Info::appRootWindow(), ! BirdtrayApp::get()->getSettings()->mIgnoreNETWMhints,
+    mWinId = findWindow(display, root, ! BirdtrayApp::get()->getSettings()->mIgnoreNETWMhints,
             BirdtrayApp::get()->getSettings()->mThunderbirdWindowMatch);
 
     Log::debug("Window ID found: %lX", mWinId );
@@ -388,9 +400,6 @@ bool WindowTools_X11::show()
 {
     if ( !checkWindow() )
         return false;
-
-    Display *display = QX11Info::display();
-    Window root = QX11Info::appRootWindow();
 
     // We are still minimizing
     if ( mHiddenStateCounter == 1 )
@@ -430,7 +439,6 @@ bool WindowTools_X11::hide()
     }
 
     // Get screen number
-    Display *display = QX11Info::display();
     long dummy;
 
     XGetWMNormalHints( display, mWinId, &mSizeHint, &dummy );
@@ -444,7 +452,7 @@ bool WindowTools_X11::hide()
 
 bool WindowTools_X11::isHidden()
 {
-    return mHiddenStateCounter == 2 && mWinId != activeWindow( QX11Info::display() );
+    return mHiddenStateCounter == 2 && mWinId != activeWindow( display );
 }
 
 bool WindowTools_X11::closeWindow()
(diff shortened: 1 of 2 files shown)


cptpcrd:
@Abestanis @gyunaev would you mind taking a look at this?

Abestanis:
@cptpcrd I'm only maintaining the translations and the Windows support. I don't have the rights to approve this, and unfortunately I also don't have a system with XWayland to test this on.

But I have some questions which might help this get moving forward:
* Why is using `XOpenDisplay` instead of `QX11Info::display` and `DefaultRootWindow` instead of `QX11Info::appRootWindow` better? Shouldn't the Qt functions just be wrappers around the xlib primitives? What's the difference?
* Looking at the description, there seems to be a subtle difference between `QX11Info::appRootWindow` (*Returns a handle for the applications root window on the given screen.*) and `DefaultRootWindow` (*return the root window for the default screen.*). Can this become problematic?
* What is the reason to introduce caching of the display an root window? Birdtray can run for a long time, what if the X server is restarted or if the root window changes? I don't know if this is an actual issue with, but maybe it would be better to revert this, we could instead just extract the getting the screen and root window into a small helper that is used everywhere.

gyunaev:
Honestly the change doesn't make much sense to me. If you're running birdtray and Thunderbird on wayland+X it should work as-is. If you're not running either under X, it won't work with this patch either. So what we are fixing here?

cptpcrd:
> Honestly the change doesn't make much sense to me. If you're running birdtray and Thunderbird on wayland+X it should work as-is. If you're not running either under X, it won't work with this patch either. So what we are fixing here?

Sorry, I didn't do a great job of explaining this in the description.

Birdtray uses Qt, and Thunderbird uses GTK. Qt and GTK each have an X11 "platform" and a Wayland "platform" that can be selected (though GDK seems to use the terminology "backend"). Thus it is entirely possible to:

1. Have Birdtray using the X11 platform (i.e. XWayland), and Thunderbird using the Wayland platform, or
2. Have Birdtray using the Wayland platform, and Thunderbird using the X11 platform (i.e. XWayland)

I only mention case (1) for completeness; it's a bit degenerate and would really require Birdtray to have full Wayland support.

My system, without any custom configuration, ended up in case (2). Thunderbird was connecting to the XWayland server, and thus _could_ have been found and hidden by Birdtray. However, since Birdtray was using Qt's Wayland platform, it didn't connect to the XWayland server and didn't even try to find Thunderbird. I worked around this by running `birdtray -platform xcb` to force Birdtray to use the X11 platform. (On various issues on this repo, people have described having to use `QT_QPA_PLATFORM=xcb` which AFAICT is equivalent to `-platform xcb`, so I suspect that this isn't just me.)

With this change, Birdtray works out-of-the-box in that situation: it continues to use the Wayland platform to render its own window, but separately connects to the XWayland server to control Thunderbird. It's not a huge deal, but I think there's some value in having Birdtray "just work" when it's straightforward to do so, which it is in this case.

To address @Abestanis's comments as well:

> @cptpcrd I'm only maintaining the translations and the Windows support. I don't have the rights to approve this, and unfortunately I also don't have a system with XWayland to test this on.

Ah, I wasn't aware.

> But I have some questions which might help this get moving forward:
> 
> * Why is using `XOpenDisplay` instead of `QX11Info::display` and `DefaultRootWindow` instead of `QX11Info::appRootWindow` better? Shouldn't the Qt functions just be wrappers around the xlib primitives? What's the difference?

When using the Wayland platform:

- `QX11Info::appRootWindow()` returns `NULL`.
- `QX11Info::display()` returns what appears to be a pointer to the _Wayland_ `wl_display`; attempting to use this as an X11 `Display` quickly segfaults.

It looks like the Qt developers assume you won't try to use most of the `QX11Info` functions except when actually using the X11 platform, which is why this PR contains an explicit check for that.

> * Looking at the description, there seems to be a subtle difference between `QX11Info::appRootWindow` (_Returns a handle for the applications root window on the given screen._) and `DefaultRootWindow` (_return the root window for the default screen._). Can this become problematic?

Perhaps. I checked the implementation for `QX11Info::appRootWindow()`, and I'm not entirely sure what happens when you don't pass an argument (in which case `screen` defaults to -1); it calls into several other APIs I'm not familiar with. However I've never seen multiple X11 screens actually be used (especially under XWayland), so I doubt it will be an issue in practice. (Note: multiple monitors are generally _not_ managed with X11 screens; there's still a single root window which stretches across all the monitors.)

> * What is the reason to introduce caching of the display an root window? Birdtray can run for a long time, what if the X server is restarted or if the root window changes? I don't know if this is an actual issue with, but maybe it would be better to revert this, we could instead just extract the getting the screen and root window into a small helper that is used everywhere.

I don't think the root window would change unless the X server restarts, and restarting the X server is fairly destructive. I don't think I've ever seen an X11 client which tries to reconnect when the server restarts; usually they just exit.

Under Wayland it's a bit different since the XWayland server can restart while the Wayland compositor keeps running, but it's still quite rare and IMO not worth the extra complexity. (Technically, there's one exception I'm aware of -- wlroots-based compositors sometimes stop the XWayland server if there are no connected clients -- but here we're keeping a connection open, so that won't happen.)

gyunaev:
Thank you for the explanation. Would it be simpler just have a check in main() for whether QX11Info::appRootWindow() returns null, and in this case self-restart with -platform xcb (i.e. exec argv[0]  -platform xcb )?

cptpcrd:
> Thank you for the explanation. Would it be simpler just have a check in main() for whether QX11Info::appRootWindow() returns null, and in this case self-restart with -platform xcb (i.e. exec argv[0] -platform xcb )?

That would work, though there's a small chance it'll break if Qt changes their platform selection logic. (Also you'd need to do some extra work to preserve other arguments like `--settings`, so it might not turn out to be simpler in the end.) IMO the change in this PR is a little more robust.

stark1tty:
I got it working over a series of periodic troubleshooting via Claude over the last year or two. Because I do not use C and Claude did the heavy lifting I do not want to push it to this repo unless given blessing by the maintainer. It works like a gem on Kubuntu 25.10 across restarts when installed locally with a full wayland desktop, Thunderbird is in the beta flatpak. I had it explain what it did in the readme, though the fix may be KDE specific.

https://github.com/stark1tty/birdtray-wayland

PR #588, opened 2024-02-19
Allow to remove account folder with Delete or Backspace keystroke

MathiusD:
When you have a large number of accounts/folder to monitor, it's quite useful to enable deletion with a keystroke on the keyboard instead of clicking a button, so this PR is made for that.

### DIFF
--- a/src/qtcomponents.h
+++ b/src/qtcomponents.h
@@ -1,0 +1,29 @@
+#ifndef QTCOMPONENTS_H
+#define QTCOMPONENTS_H
+
+#include <QKeyEvent>
+#include <QTreeView>
+
+typedef void ( * TreeViewKeyPressedEvent)(void * handle, QKeyEvent * event);
+
+/**
+ * A QTreeView that allow to define listener of keyEvent received
+ */
+class QTreeViewWithKeyEvents : public QTreeView {
+    Q_OBJECT
+
+    public:
+        explicit QTreeViewWithKeyEvents(QWidget *parent = NULL);
+
+    public slots:
+        void onKeyPressed(void * handleUsed, TreeViewKeyPressedEvent callback);
+
+    protected:
+        void keyPressEvent(QKeyEvent * event) override;
+
+    private:
+        TreeViewKeyPressedEvent mCallback;
+        void * mHandle;
+};
+
+#endif // QTCOMPONENTS_H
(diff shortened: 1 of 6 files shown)


src/qtcomponents.h
```diff
@@ -0,0 +1,29 @@
...
+    public slots:
+        void onKeyPressed(void * handleUsed, TreeViewKeyPressedEvent callback);
+
+    protected:
+        void keyPressEvent(QKeyEvent * event) override;
+
+    private:
+        TreeViewKeyPressedEvent mCallback;
+        void * mHandle;
+};
```
Abestanis, line 27: How about we just use the Qt event system instead of building our own, that way we can avoid some casts and manually storing listeners.

```suggestion
/**
 * A QTreeView that allow to define listener of keyEvent received.
 */
class QTreeViewWithKeyEvents : public QTreeView {
Q_OBJECT

public:
    explicit QTreeViewWithKeyEvents(QWidget* parent = nullptr);

signals:
    void onKeyPressed(QKeyEvent* event);

protected:
    void keyPressEvent(QKeyEvent* event) override;
};
```

with

```c++
void QTreeViewWithKeyEvents::keyPressEvent( QKeyEvent *event )
{
    emit onKeyPressed(event);
    QTreeView::keyPressEvent(event);
}
```

and

```c++
connect(treeAccounts, &QTreeViewWithKeyEvents::onKeyPressed, this, &DialogSettings::onKeyPressedOnTreeAccount);
```

Abestanis (changes requested):
I think this is a good idea, I just have an alternative proposal for the `QTreeViewWithKeyEvents` class.

Also, should we make the keyboard shortcut more discoverable? We could add a tooltip to the delete button that mentions the keyboard shortcuts.

src/qtcomponents.h
gyunaev: Looks fine with me. I mean, those are rare operations, hardly worthy optimizing, but as you wish :)

MathiusD:
> I think this is a good idea, I just have an alternative proposal for the `QTreeViewWithKeyEvents` class.
> 
> Also, should we make the keyboard shortcut more discoverable? We could add a tooltip to the delete button that mentions the keyboard shortcuts.

Yes i can had a tooltip in another PR !

> How about we just use the Qt event system instead of building our own, that way we can avoid some casts and manually storing listeners.

Yes, i have search if such signal exist in qt component and because i doesn't exist i have had own listener and i admit i've don't think to reuse signal system, i can follow your recommandation in another PR (Because it's already merged) !

PR #589, opened 2024-02-21
Refactor key press management in dialog settings

MathiusD:
Following review and comments in #588

Following PR include following modifications :
 * Rework of QTreeViewWithKeyEvents to use signal system of Qt
 * Add tooltip to present new deletion based on keystroke

### DIFF
--- a/src/dialogsettings.ui
+++ b/src/dialogsettings.ui
@@ -327,6 +327,9 @@
               </item>
               <item>
                <widget class="QPushButton" name="btnAccountRemove">
+                <property name="toolTip">
+                 <string>You can also remove folder with keystroke on Delete or Backspace key</string>
+                </property>
                 <property name="text">
                  <string>Remove</string>
                 </property>
@@ -1168,9 +1171,9 @@ p, li { white-space: pre-wrap; }
    <header>colorbutton.h</header>
   </customwidget>
   <customwidget>
-    <class>QTreeViewWithKeyEvents</class>
-    <extends>QTreeView</extends>
-    <header>qtcomponents.h</header>
+   <class>QTreeViewWithKeyEvents</class>
+   <extends>QTreeView</extends>
+   <header>qtcomponents.h</header>
   </customwidget>
  </customwidgets>
  <tabstops>
--- a/src/dialogsettings.ui
+++ b/src/dialogsettings.ui
@@ -328,7 +328,7 @@
               <item>
                <widget class="QPushButton" name="btnAccountRemove">
                 <property name="toolTip">
-                 <string>You can also remove folder with keystroke on Delete or Backspace key</string>
+                 <string>You can also remove folders with Delete or Backspace</string>
                 </property>
                 <property name="text">
                  <string>Remove</string>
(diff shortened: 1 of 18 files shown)


src/qtcomponents.h
```diff
@@ -15,15 +15,12 @@ class QTreeViewWithKeyEvents : public QTreeView {
     public:
         explicit QTreeViewWithKeyEvents(QWidget *parent = NULL);
```
Abestanis, line 16:
```suggestion
        explicit QTreeViewWithKeyEvents(QWidget *parent = nullptr);
```
and can you also please remove the `TreeViewKeyPressedEvent` since it is now unused?

src/dialogsettings.ui
```diff
@@ -327,6 +327,9 @@
               </item>
               <item>
                <widget class="QPushButton" name="btnAccountRemove">
+                <property name="toolTip">
+                 <string>You can also remove folder with keystroke on Delete or Backspace key</string>
```
Abestanis, line 331: You might not have the necessary Qt package (`Qt5LinguistTools`) installed, but since this adds a new translatable string we should add it so it adds an entry to the translations. I can do this afterwards once this is merged.

Abestanis (approved):
Awesome, thank you! ❤️

src/dialogsettings.ui
MathiusD: I hadn't seen the various translation files and had just looked at dynamic_*.ts :x

Translation added in a83c2ddd1b698996b1f5f13883af683a3492223c (For fr and en)

src/qtcomponents.h
MathiusD: Oops, I seemed to have removed it :x

NULL replaced and `TreeViewKeyPressedEvent` removed in 84c54b8bab68e4e54965c7de2d00da0f7c775a4a

src/dialogsettings.ui
Abestanis: Perfect, thanks for adding the translations!

Abestanis: Ah, and thank you for adding the other unfinished translations as well! ❤️

src/translations/main_en.ts
```diff
@@ -497,6 +497,10 @@ p, li { white-space: pre-wrap; }
 &lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-size:8pt;&quot;&gt;Thank you for your continuous support!&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>You can also remove folder with keystroke on Delete or Backspace key</source>
```
Abestanis, line 501:
```suggestion
        <source>You can also remove folders with Delete or Backspace</source>
```
I'm not a native speaker, but I think this sounds a bit more natural. What do you think?

MathiusD: Yes, it seems more natural to me too! I'll adapt the French version to be more similar

Abestanis: Perfect, thank you again! ❤️
Looks all good from my side.

PR #591, opened 2024-03-25
Hide Thunderbird after it was opened by a notification

### LINKED ISSUE #571: Becomes out of sync/unreactive when Thunderbird was opened by notification
**OS**: Gentoo Linux
**Linux Desktop Manager**: Xfce (xfce4-notifyd-0.9.1, xfwm4-4.18.0)
**Birdtray version**: 1.11.4
**Thunderbird version**: 102.15.1
**Birdtray origin**: Ebuild
**Qt version**: _NA_

**Description**
When Thunderbird generates a new mail notifcation, it includes an "Activate" button. If the button is pressed, the Thunderbird is raised. In this state, Birdtray does not hide Thunderbird again when clicking on the tray icon. TB needs to be minimized manually first.

**Expected behavior**
Birdtray should always be able to control the TB window

**To Reproduce**
1. Enable new mail notifications
2. Press "Activate" in notification once new mail arrives
3. Try to hide TB again by clicking on the tray icon

Thanks a lot!

tehw0lf:
If a notification opens the window, mHiddenStateCounter does not get updated

So we only check for _NET_WM_STATE_HIDDEN instead

Resolves #571


tehw0lf:
Could you take a look at this @gyunaev?

@jboillot you referenced this but the reference leads to nowhere. Is there a reason for this?

PR #609, opened 2024-10-08
Fix the CI

Abestanis:
This adapts the CI to the switch to Qt6. Since we also build translations and for Windows, this is build on top of and requires #608 and #607.


Abestanis:
We can either merge #607 and #608 with a broken CI, or we can just merge this which includes all changes and has a working CI.

PR #618, opened 2025-04-18
Add settings for the maximum notification font size

### LINKED ISSUE #616: Change the font size for the numbers shown
I have just installed Birdtray and it works extremely well. Thank you for all your hard work.

I was just wondering if I can reduce the size of the font used to display the number of new/unread emails? 

I realise that there is some clever code involved to make the font size fit within the icon, so, perhaps, it would be possible to reduce the size by a percentage?

Abestanis:
This exposes the maximum font setting in the advanced settings. That should close #616.

The advanced settings page

----

*This is a draft because it is build on top of the unmerged #609.*


gyunaev:
This one looks ok to me.

PR #619, opened 2025-04-18
Allow to configure an offset of the unread counter on the icon

### LINKED ISSUE #617: Move towards classic unread counter
I wonder if its possible to move towards a more conventional approach for the unread counter, as every other app does. The red dot is standard to modern applications, from mobile to desktop, and widespread on any OS system tray:
Image

the above is the standard, though some apps do it without counter, mobile style:
Image

This approach is cleaner, clearly distinguishable and doesn't require particular settings for font, color, etc.

Abestanis:
This allows to move the counter in the tray icon, **if the maximum font size is small enough that the counter can move**.
The offset is given as a positive or negative percentage, where 0% represents the center and +-100% the maximum offset.
The maximum offset takes the actual text, font size and border into account and makes sure that the number is always within the icon and never cut off.

An offset of horizontal: `100%`, vertical `-100%` and a maximum font size of `100pt` looks like this:
Tray icon with the counter at the bottom right

An offset of horizontal: `-100%`, vertical `100%` and a maximum font size of `50pt` looks like this:
Tray icon with the counter at the top left

The default is no offset, so this won't change existing setups.

I put the settings in the general tab:
image

Closes #617.

------

*This is a draft because it is build on top of the unmerged #609.*

### DIFF
--- a/src/dialogsettings.cpp
+++ b/src/dialogsettings.cpp
@@ -34,7 +34,7 @@ DialogSettings::DialogSettings( QWidget *parent)
     connect( btnNotificationIcon, &QPushButton::clicked, this, &DialogSettings::buttonChangeIcon );
     connect( btnNotificationIconUnread, &QPushButton::clicked, this, &DialogSettings::buttonChangeUnreadIcon );
     connect(borderWidthSlider, &QSlider::valueChanged, this, &DialogSettings::onBorderWidthChanged);
-connect(spinMaximumFontSize, &QSpinBox::valueChanged, this, &DialogSettings::onMaximumFontSizeChanged);
+    connect(spinMaximumFontSize, &QSpinBox::valueChanged, this, &DialogSettings::onMaximumFontSizeChanged);
     connect(spinMinimumFontSize, &QSpinBox::valueChanged, this, &DialogSettings::onMinimumFontSizeChanged);
 
     connect( treeAccounts, &QTreeView::doubleClicked, this, &DialogSettings::accountEditIndex  );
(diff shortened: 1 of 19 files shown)


gyunaev:
I wonder if those should be advanced options, as those are kinda hard to understand without reading what it is about.

Also wondering if there is a use case to have the count offset in any other position, or we should just have two options - middle-centered (default) and right bottom corner centered.

Are the numbers readable when offsets are moved? I can see it fitting numbers 1-9, but fitting something like 26 is already a challenge, and with unread emails 3-digit numbers are not uncommon.

Abestanis:
> I wonder if those should be advanced options, as those are kinda hard to understand without reading what it is about.

Ok, let's move it there, but I'll wait until #618 is merged, because merging the changes in `dialogsettings.ui` is a bit anoying.

> Also wondering if there is a use case to have the count offset in any other position, or we should just have two options

I think it's better to leave the customizations to the users. The position computation won't be much easier, so we don't even really gain anything from restricting the options.

> Are the numbers readable when offsets are moved?

Yes, it should always work, because the calculation is taking the size of the rendered text and then making sure it is always completely inside the icon.

| 10 [A] | 10 [B] | 313 [A] | 313 [B] | 1477 [A] | 1477 [B] |
| -- | -- | --- | --- | ---- | ---- | 
| Unread counter of 10 at the top left | Unread counter of 10 at the bottom right | Unread counter of 313 at the top left | Unread counter of 313 at the bottom right | Unread counter of 1477 at the top left | Unread counter of 1477 at the bottom right |
> A: horizontal: -100%, vertical: 100%, max font size: 50pt
>B: horizontal: 100%, vertical: -100%, max font size: 100pt


As you can see with 4 digits they are essentially horizontally centered.

gyunaev:
I merged it, so you're good to go.

Abestanis:
Done:
Advanced settings with new options

gyunaev:
Maybe it's time for a new release.

Abestanis:
> Maybe it's time for a new release.

Yeah, I hope the Windows installer generation still works, I haven't tested it.

PR #626, opened 2025-05-10
Avoid starting Thunderbird if we already started it

### LINKED ISSUE #625: Birdtray crashes when trying to start Thunderbird
**OS**: Linux acer 6.14.5-artix1-1 #1 SMP PREEMPT_DYNAMIC Sat, 03 May 2025 16:40:17 +0000 x86_64 GNU/Linux

**Linux Desktop Manager**: Openbox 3.6.1 Stalonetray 0.8.5
**Birdtray version**: Birdtray version 1.12.4 compiled at Apr 15 2025 18:46:50 using Qt 6.9.0.
**Thunderbird version**: 138.0 (64-bit)
**Birdtray origin**: compiled from git source
**Qt version**: Qt 6.9.0.
**Description**
Start Thunderbird
Start Birdtray
enable Settings > Hidden > Start Thunderbird when Birdtray starts & Show red X... & **Restart Thunderbird ...** & Start Thunderbird by clicking ...
Close Thunderbird from the TB window
Birdtray show a red X in the system tray
Click on Birtdtray icon
Error Reported 

Image

Click OK and Birdtray closes

**Expected behavior**
Click on Birdtray icon
Thunderbird restarts
**To Reproduce**
as above

**Screenshots**
above
**Additional context**
It seems that the problem is with the `Restart Thunderbird  if it was closed` setting. if this is left unticked then Birdtray behaves as expected.

Abestanis:
This fixes #625 by ignoring attempts to start Thunderbird if we already started it before.

Launching a new Thunderbird process while the other one was still running never really worked, because we would then delete the first process and immediately overwrite `mThunderbirdProcess` with a reference to the new process here:
https://github.com/gyunaev/birdtray/blob/6c31f6117dfc3b884e7f483adedb1c9a862187ec/src/trayicon.cpp#L696-L698

Deleting a `QProcess` stop the underlying process, and then `TrayIcon::tbProcessFinished` would be called, which then deleted and stopped the new process because `mThunderbirdProcess` already points to the new process:
https://github.com/gyunaev/birdtray/blob/6c31f6117dfc3b884e7f483adedb1c9a862187ec/src/trayicon.cpp#L738

I tested this locally on Windows and confirmed that I couldn't reproduce the bug anymore, but I'm marking this as a draft while waiting for a verification from the reporter that this really fixes the problem for them.

### DIFF
--- a/src/trayicon.cpp
+++ b/src/trayicon.cpp
@@ -27,7 +27,7 @@ TrayIcon::TrayIcon(bool showSettings)
 
     mMenuShowHideThunderbird = 0;
     mMenuIgnoreUnreads = 0;
-    mThunderbirdProcess = 0;
+    mThunderbirdProcess = nullptr;
 #ifdef Q_OS_WIN
     mThunderbirdUpdaterProcess = ProcessHandle::create("updater.exe");
     connect( mThunderbirdUpdaterProcess, &ProcessHandle::finished,
@@ -707,7 +707,7 @@ void TrayIcon::startThunderbird()
     mThunderbirdProcess->start(executable, args);
 }
 
-void TrayIcon::tbProcessError(QProcess::ProcessError )
+void TrayIcon::tbProcessError(QProcess::ProcessError error)
 {
 #ifdef Q_OS_WIN
     if (mThunderbirdUpdaterProcess->attach() == AttachResult::SUCCESS)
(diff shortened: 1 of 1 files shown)


Abestanis:
Another symptom of this bug and the reason of the crash described in #625:
When the `TrayIcon::tbProcessFinished` of the first is called, it deletes and nulls `mThunderbirdProcess`, which then kills the second process. This in turn causes a call to `TrayIcon::tbProcessError`, where we then attempted to use the nulled `mThunderbirdProcess`, which then segfaulted.

Abestanis:
The reporter needs some time to test if this actually fixes their problem, but I think these are good anyways. And I'm pretty confident that they will fix the problem, since I replicated it on my machine and couldn't replicate it anymore with this patch.

AndyM48:
I think I am confused.

I cloned master into a new directory and went to edit trayicon.cpp. The lines above are already in the file. Should I remove them?

PS. Compiled OK with cmake 4

AndyM48:
Just to confirm, installed master and everything works as expected

Thanks for your help

Abestanis:
> Just to confirm, installed master and everything works as expected

Perfect, thanks for all of the information and verifying the fix.

